PR #10, opened 2023-10-18
feat: create logic to generate commands dynamically in master plugin

FlorentLvr:
## JSON command

image

## Detailed

**- name :** Notebook with extension 

image

**- description:**  Notebook description replacing the word "notebook" by "command" 

image

**- command parameters** = variables used in the cell below the cell tagged "#parameters"

image

**- parameter description**: description in setup variable cell

image

**- parameter default**: default below "setup variable" cell

image


FlorentLvr:
@Dr0p42, this feature will be helpful to create a plugin by engine as it can dynamically create commands from an input directory to find all files with commands

FlorentLvr:
generate avatar by command based on image on the first cell of the notebook:
image

FlorentLvr:
@Dr0p42, I added the plugin on each engines! Feel free to merge :)

Dr0p42:
I am not fully convinced about the way the commands are being generated automatically. Are we also planning to have a way to:
- Exclude some notebooks from being automatically added? (Removing `#command` should do it right)
- Having an explicit way to create a command by calling a function and specifying all parameters?

Here for example the description of the command might be a bit confusing no?

image

What do you think?

PR #180, opened 2024-07-23
fix: Make ABI detect that there is a drift between conda.yml and the installed conda env

Dr0p42:
This pull request add a feature in the Makefile and the setup.ipynb that will update the conda env if the conda.yml changed.


Dr0p42:
@FlorentLvr can you review and merge please?

PR #353, opened 2025-02-10
fix: image gen location

jravenel:
The image was still save in the previous data folder and not in storage


FlorentLvr:
@jravenel, storage service must be used in integration and image_dir must be passed as configuration and not inside the function.

FlorentLvr:
@jravenel i fixed it by using storage services and naas integration to create asset

PR #362, opened 2025-02-25
[WIP] feat: Add test command to terminal agent

jravenel:
@Dr0p42 
I added the mock up of the test we talked about, now we need to make sure to add the github "how many issues open" answer correctly with split Issue and PR. 

╭───────────────────────────────────────╮
│ Welcome to ABI Super Assistant        │
│                                       │
│ Available commands:                   │
│ - 'exit' to end the conversation      │
│ - 'reset' to start a new conversation │
│ - 'help' to see these commands again  │
│ - 'test' to run tests and coverage    │
╰───────────────────────────────────────╯

───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

You: test

Running tests...
Running unit tests...
Running integration tests...
Calculating coverage...

Test Results:
Tests: 198/200 passed
Coverage: 89.5%

You: exit

### DIFF
--- a/src/core/apps/terminal_agent/main.py
+++ b/src/core/apps/terminal_agent/main.py
@@ -1,4 +1,9 @@
-from src.core.apps.terminal_agent.terminal_style import clear_screen, print_welcome_message, print_divider, get_user_input, print_tool_usage, print_agent_response, print_tool_response, print_image
+from src.core.apps.terminal_agent.terminal_style import (
+    clear_screen, print_welcome_message, print_divider, 
+    get_user_input, print_tool_usage, print_agent_response, 
+    print_tool_response, print_image
+)
+from test.test_runner import run_mock_tests
 from abi.services.agent.Agent import Agent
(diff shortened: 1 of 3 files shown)


Dr0p42:
Should the tests be executed from the agent? It's not usual, I thought we would run it from the Makefile, like `make tests`. 

What is the goal of the test itself? If I recall we want to test, if an Agent + a tool is providing the correct output right? So the execution should be something like:
- Create a tool, mocking Github for example.
- Create the Agent with the tool.
- Send a specific prompt "How many issues and pull requests are being opened"
- Get the result.
- Call an llm, with the context:
 - Explain the goal of the test.
 - Provide the number of issues and PR
 - Provide the Agent output.
 - Ask if the response from the agent is correct or not. We need a Score or just a boolean to know if it's good.

Is that the target? If yes it should allow us to test different tools/agent for specific prompts. But for each test we need to have access to the ground truth.

PR #433, opened 2025-04-29
[WIP] docs: v0 ontology with Organization as example

FlorentLvr:
Create docs for ontology classes and relations


FlorentLvr:
@jravenel v0 push with design pattern

PR #459, opened 2025-05-26
clean src/core with uv ruff mypy check

FlorentLvr:
- Remove cyber ontologies and documentation
- clean src/core with uv ruff mypy check
- Refacto ontology agent


FlorentLvr:
@Dr0p42, I started to fix errors. There are still "Found 117 errors in 20 files (checked 68 source files)". Some are coming from "src/core/modules/ontology/pipelines/AddCommercialOrganizationPipeline.py:224: error: Signature of "as_api" incompatible with supertype "Expose"  [override]" could you please have a look? I commented again the Makefile to push the files i already fixed

PR #515, opened 2025-08-10
Research tools io

jravenel:
First iteration done 
Screenshot 2025-08-10 at 11 12 18
Screenshot 2025-08-10 at 11 12 36
Screenshot 2025-08-10 at 11 12 48

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,6 @@
 {
     "python.analysis.extraPaths": [
         "./lib"
-    ]
+    ],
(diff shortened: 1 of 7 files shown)


FlorentLvr:
@jravenel, i am not sure this FilesystemTool should be in `/lib`. From what I understand from your code, this tool can only work in dev and as only access to files from your local. It should be implemented as tools in a dedicated module using src.services to perform actions on object. For the access to the code base, maybe it will be better to have in a vector DB. 
What do you think @Dr0p42 ?

PR #605, opened 2025-09-22
feat: add airgap support for intent mapping

jravenel:
- Add dimension configuration for airgap mode (768 vs 1536)
- Support different embedding dimensions based on AI_MODE
- Maintain compatibility with both cloud and airgap embeddings
- Fix import ordering in Embeddings.py

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 .DEFAULT_GOAL := default
 
 # Default target with help display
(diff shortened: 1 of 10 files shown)


jravenel:
@Dr0p42 I think it's good to go. 
I checked airgap with qwen, cloud with openai api key, added test cov

PR #619, opened 2025-09-30
fix embeddings dimensions

jravenel:
- Separate cache directories for different embedding providers
- openai_1536/ for OpenAI embeddings (1536 dimensions)
- airgap_768/ for local airgap embeddings (768 dimensions)
- Prevents dimension conflicts when switching between providers
- Maintains performance benefits for each provider independently
- Future-proof for additional embedding providers

### DIFF
--- a/docs/troubleshooting/README.md
+++ b/docs/troubleshooting/README.md
@@ -1,0 +1,42 @@
+# Troubleshooting Guide
+
+Quick fixes for common ABI issues.
+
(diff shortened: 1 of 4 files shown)


jravenel:
@Dr0p42 I changed the logic and it works now. 
Test performed: 

- my .env = airgap 
- I change to = cloud
- it adapts and start abi

PR #634, opened 2025-10-06
Remove data examples from storage

### LINKED ISSUE #633: Clean data examples in storage
We don't need them anymore.

jravenel:
- Delete all example directories from datastore, triplestore, and vectorstore
- Remove .gitignore rules for tracking example files
- Simplify storage to focus on actual data only


jravenel:
as discussed today, rm the examples here

PR #816, opened 2026-03-29
feat(nexus): support Ollama in Docker via OLLAMA_HOST, integrate OpenCode in Lab, improve agents list display

jravenel:
## What changed

**Ollama in Docker (the core bug fix)**
- Nexus containers were resolving `localhost:11434` to themselves — not the host. Fixed by reading an `OLLAMA_HOST` env var (default: `http://localhost:11434`). Set it to `http://host.docker.internal:11434` in `docker-compose.override.yml` on macOS/Windows Docker Desktop.
- Added an explicit `provider='ollama'` branch in `_resolve_provider()` (`chat.py`) — Ollama agents were silently falling through to the cloud provider block.

**Model registry expansion**
- Added `qwen2.5-coder:7b`, `qwen2.5-coder:14b`, `deepseek-r1:8b`, `phi:latest` as local Ollama options.

**Nexus Lab — OpenCode AI coding assistant**
- Integrated `opencode` as an AI agent in the Lab editor: sends messages via the `opencode` API (SSE streaming), auto-approves `bash`/`write`/`edit`/`delete`/`patch` tool calls, and includes a polling fallback for when SSE misses the `session.idle` completion event.

**Nexus Lab — file refresh**
- Added a manual refresh button (toolbar) to reload any open file from disk.
- Active file auto-reloads after the agent finishes editing it.

**Nexus Lab — integrated terminal**
- Added an `xterm.js` terminal connected to a PTY WebSocket backend (`lab_terminal.py`).

**Cleanup**
- Fixed Ruff `I001` import-sort errors in `lab_fs.py` and `lab_git.py` — `naas_abi` is third-party so it belongs in the same block as `fastapi`/`pydantic`, sorted alphabetically.
- Moved `@xterm/xterm/css/xterm.css` import from `globals.css` (where it broke Next.js CSS processing) into `terminal-pane.tsx`.
- Exclude `docker-compose.override.yml` from Git — it's user-local.


jravenel:
Need to break this PR into multiple pieces before merge.

PR #948, opened 2026-05-13
fix(local-dev): make local stack work out of the box

jravenel:
## Summary

Makes `abi deploy local` + `abi start` work out of the box, end-to-end:

**Stack lifecycle**
- Stack no longer crashes on startup without SMTP — engine's `services.email` block now wires a filesystem adapter for local dev; Nexus auth honors that adapter instead of duplicating SMTP wiring.
- `GITHUB_ACCESS_TOKEN` no longer hard-crashes when the GitHub module is disabled — removed from `config.local.yaml` (still available via marketplace).
- `abi start` opens Nexus (port 3042) with the admin email pre-filled, not a generic services portal.

**Auth**
- Login page shows password form by default for local dev (`auth_password_enabled: true` in `config.local.yaml`).
- `login()` action implemented in the auth store; password form now actually authenticates.
- Logout button clears tokens and redirects.
- `auth/config` result cached in `sessionStorage` so the login page renders instantly on revisit instead of blocking on a round-trip.
- Cookie `nexus-auth-flag` set with `SameSite=Lax` to match middleware-set cookies.

**Email**
- Magic-link email goes through `engine.services.email` (FastAPI `Depends`), pulled from `app.state.email_service` set by `ABIModule`. Adapter swap (filesystem ↔ SMTP) is now a single config-yaml change, not a code change. SMTP transport settings removed from Nexus config; only `email_from_address` / `email_from_name` remain Nexus-local.

**Workspace routing**
- Cookie + env-var fallback + `'primary'` default replaces the hardcoded `workspace-nexus` slug.
- `DEFAULT_WORKSPACE` env wired into `nexus-web` in both root and CLI-template docker-compose.
- Post-login workspace ID returned by `login()` is now picked from a stable sort (`created_at`, then `slug`) instead of arbitrary array order.

**CLI**
- `abi deploy local` templates now match the dev-repo setup: admin credentials, `ABI_API_KEY`, `DEFAULT_WORKSPACE` are seeded; `OPENAI_API_KEY` / `NAAS_API_KEY` are left absent so the CLI prompts the user for them instead of treating `""` as an answer.
- `stack.py` reads `NEXUS_WEB_PORT` / `NEXUS_USER_ADMIN_EMAIL` through `EngineConfiguration.load_configuration().services.secret.load()` (lru-cached) instead of a hand-rolled `.env` parser. Emails are URL-encoded so addresses with `+` or `&` work.

**Cleanup (out of scope of "make local dev work," but loud orphans removed since this PR was already touching everything)**
- Removed: legacy `seed.py`, `demo/` directory, `db-seed` Makefile targets, `auto_seed_demo_data` setting in both `naas_abi/__init__.py` and `apps/api/app/core/config.py` (call site was already removed).
- Removed: Matrix/Element commented-out service blocks from root + CLI-template docker-compose, the `element-config.json` files, `synapse/` bootstrap, and matrix references in the service-portal index template + CLI service catalog. Never enabled by default; dead weight.
- Removed: root `.env.example`. The CLI templates (`abi deploy local`) are the canonical bootstrap; `.env.example` was a parallel source of truth that drifted. Quickstart docs updated to inline the minimum keys.
- `NEXT_PUBLIC_API_URL` now points at the public host (browser-reachable) instead of the internal Docker hostname; port corrected from 8080 to 9879.
- `NEXUS_USER_ADMIN_PASSWORD` in `.env` now actually used by the seed code (previously silently ignored).
- `show_powered_by` defaults to `false` so the "Powered by NEXUS" badge doesn't leak when the config API is unreachable.
- `the-stack.mdx` — Engine renamed to Context Engine.
- README quick start and How It Works rewritten against the actual architecture; UI screenshots added.

## Commit map

1. `fix(cli): align deploy templates + drop dead .env.example` — CLI deploy templates, `stack.py` engine-secret-service refactor, removal of placeholder values.
2. `chore: remove unused matrix/element services and stale comment refs` — matrix/element/synapse cleanup, quickstart doc update.
3. `fix(auth): route magic-link email through engine email service` — option A: plumb `engine.services.email` onto `app.state`, refactor `_send_magic_link_email`, strip duplicated `smtp_*` settings, drop orphan `auto_seed_demo_data`.
4. `fix(web): cookie SameSite, stable workspace ordering, login render` — small auth/UX polish items from review.

---

## Test plan

- [ ] `uv tool install naas-abi-cli --force --upgrade && abi new project foo && cd foo && abi deploy local && abi start`
- [ ] Stack comes up healthy (all containers green, no unhealthy `abi` container)
- [ ] Browser opens to `http://localhost:3042/auth/login` with email pre-filled
- [ ] Login with `admin@example.com` / `Admin1234!` succeeds and lands on the workspace
- [ ] Logout button works and returns to login page
- [ ] Reload `/auth/login` — page renders immediately (no spinner), password form still visible (sessionStorage hit)
- [ ] No "Powered by NEXUS" visible anywhere in the UI
- [ ] `http://localhost:9879/docs` shows the API docs
- [ ] Magic-link path: stop the stack, set `services.email.email_adapter.adapter: smtp` in `config.local.yaml` with a real mailcatcher, restart, request magic link — message lands in mailcatcher


FlorentRvl:
@jravenel, why did you remove suggestions in AbiAgent.py ? 

You can change it you want but we need to have one suggestions at least so we can test this feature with abi.

image

jravenel:
> @jravenel, why did you remove suggestions in AbiAgent.py ?
> 
> You can change it you want but we need to have one suggestions at least so we can test this feature with abi.
> 
> image

@FlorentRvl bcause it's just a placeholder for now, but I will create a new issue as we have to improve this feature with description under le placeholder, capability to grey boxes (not clickable, coming soon) and abiity to do CTA on other parts of the sidebar, like: go to Apps

PR #1090, opened 2026-07-25
fix(nexus): use OpenRouter for voice transcription when available

### LINKED ISSUE #1089: Nexus voice transcription fails with OpenRouter keys (401)
## Current situation
Nexus voice transcription calls `api.openai.com` with `OPENAI_API_KEY`.

Deployments that use an OpenRouter key (`sk-or-...`) for chat get:

`OpenAI transcription failed (401)`

## Target situation
Voice transcription works with the same OpenRouter key used for chat.

Use OpenRouter STT (`/api/v1/audio/transcriptions`) when an OpenRouter key is present. Keep native OpenAI as an optional override.

jravenel:
## Summary
- Route Nexus `/transcribe` through OpenRouter STT when an OpenRouter key is present.
- Keep native OpenAI transcription as an explicit override via `OPENAI_TRANSCRIBE_API_KEY`.
- Load OpenRouter `HTTP-Referer` / `X-Title` from Nexus settings (`frontend_url`, `magic_link_email_app_name`) instead of hard-coding a product name. Also applied to `provider_runtime`.

## Test plan
- [ ] With only `OPENROUTER_API_KEY` set, POST audio to `/transcribe` returns 200 and text
- [ ] With `OPENAI_TRANSCRIBE_API_KEY` set to a native OpenAI key, requests still go to `api.openai.com`
- [ ] OpenRouter key shaped `sk-or-...` is never sent to `api.openai.com`
- [ ] OpenRouter attribution headers come from config/settings, not a hard-coded brand

Fixes #1089

### DIFF
--- a/libs/naas-abi/naas_abi/apps/nexus/apps/api/app/api/endpoints/transcribe.py
+++ b/libs/naas-abi/naas_abi/apps/nexus/apps/api/app/api/endpoints/transcribe.py
@@ -1,4 +1,4 @@
-"""Audio transcription endpoint backed by OpenAI."""
+"""Audio transcription endpoint (OpenRouter or native OpenAI)."""
 
 from __future__ import annotations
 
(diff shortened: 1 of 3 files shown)


jravenel:
@Dr0p42 @FlorentRvl could you review this when you have a moment?

OpenRouter STT routing for Nexus `/transcribe` (fixes 401s when chat uses an `sk-or-...` key), plus OpenRouter attribution headers from settings/config instead of a hard-coded product name.

FlorentRvl:
> @Dr0p42 @FlorentRvl could you review this when you have a moment?
> 
> OpenRouter STT routing for Nexus `/transcribe` (fixes 401s when chat uses an `sk-or-...` key), plus OpenRouter attribution headers from settings/config instead of a hard-coded product name.


TY for your contribution. This should work but i think we should use the ModelRegistry to register the default voice model to avoid hardcode the transcribe code. The might needs to have a voice "services" with adapters so we can switch models in transcribe without changing the code. 
What do you think @Dr0p42 ?

jravenel:
@FlorentRvl @Dr0p42 Agree on the direction.

Short term: this PR should stay a 401 fix (route /transcribe through OpenRouter when we have an sk-or- key, keep native OpenAI as an explicit override, load attribution from settings). ModelRegistry today is chat + embedding only: no voice/STT model type, no default voice model, no voice provider factory. Wiring defaults through the registry here would mean expanding the core contract inside a bugfix.

Follow-up (happy to open an issue + ADR after merge): a voice service with adapters (OpenRouter / native OpenAI, and later others) so /transcribe and /speech resolve backend + model without hardcoded IDs. Same pattern should cover #1092 TTS.

If that split works for you, let's approve this PR as-is and we will track the voice-service work separately?

jravenel:
Tracking Florent's ModelRegistry / voice-service suggestion here: #1101

PR #1094, opened 2026-07-25
fix(nexus): remap reserved create-skill slugs in skill drafts

### LINKED ISSUE #1093: Skill Creator rejects drafts that use /create-skill as the slug
## Current situation
`/create-skill <description>` asks the agent to draft a skill. The model often sets `slug` to `create-skill`.

Saving fails with: `'/create-skill' is a reserved command`.

## Target situation
Skill drafts never keep reserved builtins (`skills`, `create-skill`) as the slug.

Derive a task slug from the skill name, show it in the Skill Creator card, and allow the user to edit it before save.

jravenel:
## Summary
- Remap reserved skill slugs (`create-skill`, `skills`) to a name-derived slug on create.
- Skill Creator card shows an editable slug and warns when a reserved value was remapped.
- Prompt instructs the model never to use reserved builtins as the skill slug.

Closes #1093

## Test plan
- [x] Run `/create-skill that summarizes meetings` and confirm the draft slug is not `create-skill`
- [x] If a draft still has `create-skill`, the card remaps it and Save skill succeeds
- [x] Duplicate non-reserved slugs still fail validation

### DIFF
--- a/libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/chat/service.py
+++ b/libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/chat/service.py
@@ -122,6 +122,10 @@ _CREATE_SKILL_INSTRUCTIONS = (
     "exactly these keys: name, slug, description, prompt. Briefly explain your choices outside "
     "the code block. Do not attempt to save it yourself — the user reviews and saves it from "
     "the UI with one click.\n"
+    "Critical: the slug is the chat command users will type later (e.g. `/weekly-report`). "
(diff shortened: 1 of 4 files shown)


jravenel:
@Dr0p42 @FlorentRvl could you review this when you have a moment?

Fixes Skill Creator failing with `'/create-skill' is a reserved command` when the model echoes the builtin slash command as the skill slug.

FlorentRvl:
> @Dr0p42 @FlorentRvl could you review this when you have a moment?
> 
> Fixes Skill Creator failing with `'/create-skill' is a reserved command` when the model echoes the builtin slash command as the skill slug.

TY for your contribution @jravenel ! Good to merge!

PR #1113, opened 2026-07-29
fix: prompt for new module CLI #1112

### LINKED ISSUE #1112: Fix new module CLI output to match new configuration file structure
# Description

When currently using the `abi new module`, the output prompt to be added in the YAML configuration file to enable the module seems to be in the former file syntax.

```bash
(bob) alexis@FR012TOURALEX:~/proj/github/bob$ abi new module "Assurance Veille Reglementaire" src/internal/
[abi] ABI project detected. Re-running in local project environment.
      project: /home/alexis/proj/github/bob
      versions: system=2.4.0 | project=2.9.4

Module 'assurance-veille-reglementaire' has been created at:
  src/internal/assurance_veille_reglementaire

To enable this module, add the following to your config.yaml:

modules:
  - path: src/internal/assurance_veille_reglementaire
    enabled: true
```

# Task

- [ ] Fix the output message to match the new configuration structure

```yaml
modules:
  - module: internal.assurance_veille_reglementaire
    enabled: true
```

AlexTorxFM:
Closes #1112

# Content

- Update the prompt display to match new configuration file syntax

# Caveats

The module name displayed wthin the help message depends on the current working directory and/or whether the user uses a relative or absoute path as module directory. I added a prompt message to ask the user to be careful about what would the be the expected namespace.

### DIFF
--- a/libs/naas-abi-cli/naas_abi_cli/cli/new/module.py
+++ b/libs/naas-abi-cli/naas_abi_cli/cli/new/module.py
@@ -89,8 +89,15 @@ def new_module(module_name: str, module_path: str = ".", quiet: bool = False):
     )
 
     if not quiet:
+        # Configuration file requires an importable module namespace rather than an
+        # actual file path on the filesystem (so we need to replace '/' with '.')
         print(f"\nModule '{module_name}' has been created at:\n  {module_path}\n")
         print("To enable this module, add the following to your config.yaml:\n")
         print("modules:")
-        print(f"  - module: {module_path}")
+        print(f"  - module: {module_path.replace("/", ".")})
         print("    enabled: true\n")
+        print(
+            "Note: please check the module import namespace root level that might need "
+            "to be updated depending on your actual project structure.\n"
+        )
+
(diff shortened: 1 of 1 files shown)


Dr0p42:
Heads-up on the red CI here 👇 — both `check (3.11)` and `check (3.12)` fail at `make check` → `check-core`, with ruff reporting 6 `invalid-syntax` errors from this line:

```python
print(f"  - module: {module_path.replace("/", ".")})
```

Two issues on it:
1. The f-string is **unterminated** — the closing `"` before `)` is missing (invalid in every Python version).
2. It reuses double quotes inside its own replacement field. Nested same-type quotes require PEP 701 (3.12+), and ruff here is pinned to `target-version = "py311"`, so it's rejected even on the 3.12 job.

There's also a latent bug on the same line: `module_path` is absolutised when the default `.` is used (`os.path.join(os.getcwd(), ...)` at line 30), so `.replace("/", ".")` prints a namespace like `.Users.me.dev.project.my_module` rather than `my_module`.

Suggested replacement:

```python
module_namespace = os.path.relpath(module_path, os.getcwd()).replace(os.sep, ".")
...
print(f"  - module: {module_namespace}")
```

Since this PR is from a fork with `maintainerCanModify: false`, I couldn't push the fix onto your branch — so I put it in #1130, which carries your two commits unchanged plus one fix commit and a `module_test.py` (this branch wasn't executed by any test, which is how the syntax error got through). Feel free to either apply the snippet above here, or close this in favour of #1130 — whichever you prefer.

For the record, `- module:` is the right key: `ModuleConfig` (`EngineConfiguration.py:240`) accepts `module` as the dotted namespace with `path` as the legacy form. 👍

AlexTorxFM:
Closing in favor of #1130 

Latest fix commit was not pushed yet, so agree with the review 😄 Thanks for the review

PR #1125, opened 2026-07-29
feat(nexus): create-on-invite for org and workspace members

jravenel:
## Summary
- Org/workspace member invite now **creates the user when missing**, adds membership, and emails OTP / magic-link sign-in (reuses the #1124 challenge path).
- Optional `workspace_id` on org invite also attaches a workspace in one call.
- Same capability via Users UI copy, Abi `invite_*` admin tools, and `abi user invite` / `abi org members invite` / `abi workspace members add`.

## Stacking
- **Depends on:** #1119 (CLI), #1124 (email OTP), #1122 (Users invite UI), soft #1128 (admin tools invite hunks).
- Also includes #1117 via #1122.
- Unique slice is tip commit `feat(nexus): create-on-invite...` (G). Dep commits are included so the branch builds; merge deps first, then this PR (or rebase after deps land).
- Megamerge / compare / pane-send stripped. Those live in #1123 / #1128.

## Test plan
- [ ] Unit: `auth/service_test.py` ensure_user_for_invite + existing org/auth/admin tool tests
- [ ] API: `POST /api/organizations/{org}/members/invite` with a never-seen email returns member + `user_created: true` and sends sign-in email
- [ ] Optional: same call with `workspace_id` adds workspace membership
- [ ] UI Add User on org Users: new email succeeds; copy no longer says must already exist
- [ ] Abi tool + CLI invite paths work
- [ ] Invitee can log in via OTP or magic link

PR #1124, opened 2026-07-29
feat(nexus): email OTP sign-in alongside magic link

jravenel:
## Summary
- Add a 6-digit email OTP on the existing passwordless magic-link challenge so users can type a code instead of click-link + confirm.
- Keep magic link as email fallback; same token row, expiry, one-time use, hashed secrets, rate limits, and max OTP attempts.
- Login UI (when password auth is off): send code → enter OTP → session; magic-link page unchanged.

## Stacking
- Wave 1 independent off `main`.
- **P7** (#1125 create-on-invite) depends on this for invite sign-in email / OTP path.

## Test plan
- [ ] `cd libs/naas-abi/naas_abi/apps/nexus/apps/api && uv run pytest app/services/auth/service_test.py app/services/auth/adapters/primary/auth__primary_adapter__FastAPI_test.py -v`
- [ ] With password auth disabled: request sign-in, enter emailed 6-digit code, land in workspace
- [ ] Confirm magic link in the same email still works (confirm page)
- [ ] Wrong code increments attempts; after `otp_max_attempts` code is invalidated
- [ ] Replay of a used code returns 400
- [ ] Migration `0038_add_magic_link_otp.sql` applies cleanly on API startup

PR #1123, opened 2026-07-29
feat(nexus-web): dual-chat compare pane + pane-send fixes

jravenel:
## Summary
- Replace the right AI mini-widget with a real `ChatInterface` surface (`surface=\"pane\"`): bubbles, tool-call UI, uploads, composer, agent selector, independent conversation.
- Cursor-style chrome: open-chat tabs (plus new/history/close); Columns2 \"Send to both\" when the pane is open.
- Left section panel and right chat pane are drag-resizable (widths persisted).
- Pane-send fixes from zen tip: restore send when conversation id is orphaned; stop no-op across workspace switches.

## Stacking
- **Depends on:** #1128 (P5 Abi pane + admin tools). Base branch is `feat/abi-pane-nexus-admin-tools`.
- Slice E + tip pane-send only. Stripped Users/OTP/create-on-invite/CLI bloat from earlier megabranch tip.

## Test plan
- [ ] Open Sparkles / Cmd-K: right pane is full chat UI with tab strip
- [ ] History (clock) opens recent chats; Plus starts New chat; close focuses neighbor
- [ ] Send in right pane works (new chat + after workspace switch)
- [ ] Pick different agents left vs right; Columns2 sends same prompt to both
- [ ] Drag left/right edges: widths persist after reload
- [ ] Closing pane does not clobber main chat conversation/URL

PR #1122, opened 2026-07-29
feat(nexus): wire org Users UI to member invite API

jravenel:
## Summary
- Replace mock Alice/Bob/Carol + \"coming soon\" on org settings **Users** with live `GET/POST /api/organizations/{orgId}/members` (invite existing user by email + role), same APIs as CLI #1119.
- Enforce owner/admin in UI (disable Add User); API remains the ABAC source of truth (`403`).
- Align workspace **Members** invite UX (errors, loading, admin gate) and workspace org Users page.
- Document the shared UI/CLI/agent capability surface in Nexus `AGENTS.md`. Agent tool wrappers tracked in #1121.

## Stacking
- **Depends on:** #1117 (Users framing). Base branch is `fix/nexus-org-users-not-admins`.
- Merge #1117 first, then this PR.

## Test plan
- [ ] `/organizations/{orgId}/settings/users` lists real members (no demo names)
- [ ] Owner/admin: **Add User** invites existing email; list refreshes; unknown email shows API error
- [ ] Non-admin: Add User disabled; direct API invite returns 403
- [ ] Workspace Settings → Members: invite shows inline errors; Invite disabled for non-admins
- [ ] `pnpm exec tsc --noEmit` and org-settings route vitest pass
- [ ] Live-test on zen after ABI pin

PR #1126, opened 2026-07-30
chore(zen): zen-next-abi-20260731

jravenel:
## Summary

Single review vehicle for everything ahead of `jupyter-naas/abi` `main` that is live on Zen, consolidated on branch `zen-next-abi-20260730`.

- **Tip SHA:** `26b1374fd` (`chore(slides): ship three preset seeds with minimal-light default`)
- **Zen pin:** jupyter-naas/zen pins submodule `.abi` @ `26b1374fd` (zen tip `54525f2` on `zen-next-abi-20260730`)
- **Ahead of `main`:** do not force-push `main`

**Merge care:** Review and land via this PR (or an agreed merge strategy). Prefer reviewing the tip as a unit; squash/rebase/merge sequencing may need care because this branch carries stacked product work. Do not merge until Jeremy asks.

## Full scope on tip (live on Zen)

- **Auth:** Email OTP sign-in alongside magic link
- **Invites:** Create-on-invite for org and workspace members
- **Chat pane:** Dual-chat compare, Cursor-style tabs/history, pane-send fixes (orphan conversation / workspace switch), Abi default on right pane
- **Org admin:** Users (not Admins) framing; Users invite UI; Workspaces column multi-select assign/remove; Roles feature matrix (owner/admin/member/viewer)
- **Abi agent:** Default AI pane to Abi with org/workspace admin tools; Claude Sonnet / context-safe tooling coordination with Zen pin
- **CLI:** Workspace and user admin commands
- **Coding:** Forgejo org `abi` + Coder template seeding
- **Slides MVP:** Forgejo storage, Monaco editor, File/View menu bar (File → **Save**), presentation file tree, Preview|Code toggle, Abi pane (⌘K)
- **Slides Abi loop (MVP):** Required per-slug Coder runtime (`abi-slides`) with persisted sidecar binding; Abi tools prefer Coder workspace FS and fall back to Forgejo; open-deck slug/title/mode injected into pane context so Abi does not ask which deck
- **Slides preview:** Sidecar-first Preview refresh after Abi edits; `replace_in_slides_deck` entity-aware (`&`/`&amp;`, U+2014/`&mdash;`/`&#8212;`) for cover titles and subtitles (`cover_subtitle_updated`)
- **Slides templates:** Three gallery presets: Minimal Light (`minimal-light-v1`, default), Pitch Dark (`pitch-dark-v1`), Executive (`executive-v1`)
- **Platform status footer:** Shell-wide thin footer (User / Business workspace / Repo / Branch / Code workspace + Saved/Unsaved; Refresh + API). Code workspace opens Coder dashboard when online. Navbar has no branch/API chrome; footer owns that truth. UX note: Zen `docs/ux/business-vs-code-workspace.md`
- **Pane export:** AI pane transcript export via overflow menu

## Superseded split PRs (closed; review here)

Do not review or merge these separately; their intent is on this tip:

- #1117 Users not Admins
- #1119 CLI workspace/user admin
- #1122 Org Users invite UI
- #1123 Compare chat + pane-send
- #1124 Email OTP
- #1125 Create-on-invite
- #1128 Abi pane + admin tools

## Intentionally left open (not this vehicle)

- #1070 Desktop app
- #964 Code IDE section (separate track unless later folded)
- Florent / Maxime / Alexis PRs and other unrelated work

## Test plan

- [ ] zen.naas.ai right pane send works (new chat + after workspace switch)
- [ ] Compare / tabs / history still work
- [ ] Email OTP sign-in still works
- [ ] Create-on-invite still works
- [ ] Org Settings > Users: invite + Workspaces assign/remove
- [ ] Org Settings > Roles: feature matrix loads; Save toggles apply (persist story as noted in product)
- [ ] Slides: open deck → runtime ensure (banner if Coder down) → Abi pane chip shows open path → ask "just replace the title" without Abi asking which deck
- [ ] Slides: after Abi write, Preview refreshes from sidecar; cover `<h1>` / subtitle updates (entity-aware replace including `&mdash;`)
- [ ] Slides: New Presentation gallery lists Minimal Light, Pitch Dark, and Executive; create seeds the selected preset (default Minimal Light)
- [ ] Shell footer Code workspace link opens Coder UI for that env; Slides shows Saved / Unsaved changes
- [ ] Shell footer on Chat/Code/Slides: User + Business workspace + API; Repo/Branch/Code workspace filled on Code/Slides; navbar has no `main`/API chrome
- [ ] Pane transcript export
- [ ] Coding seeds / Forgejo `abi` org as expected on Zen
- [ ] Agree merge strategy before landing on `main`


libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
```diff
@@ -0,0 +1,1289 @@
...
+    workspace_id: str,
+    request: Request,
+    current_user: User = Depends(get_current_user_required),
+) -> list[ProjectResponse]:
+    await require_workspace_access(current_user.id, workspace_id)
+    sc = _get_source_control(request)
+    repo_id = _repo_id()
+
+    def _list() -> list[ProjectResponse]:
+        branches = sc.list_branches(repo_id=repo_id)
```
Dr0p42, line 813: [P0] Scope slide projects to the owning workspace. `workspace_id` is currently used only to prove that the caller belongs to *some* workspace; this then enumerates every `slides/*` branch in the single global repo, and the get/put/history/tree routes likewise address projects only by global slug. A member of workspace A can therefore list, read, or overwrite workspace B's deck by supplying B's slug, and slug collisions prevent separate tenants from using the same name. Persist the owning `workspace_id` and enforce it on every operation (or namespace branch/path by workspace). Also avoid using repo-wide `add_collaborator(..., permission="write")` as per-deck authorization, since that grants the Forgejo user write access beyond this deck.

libs/naas-abi/naas_abi/apps/nexus/apps/web/src/app/workspace/[workspaceId]/slides/[slug]/page.tsx
```diff
@@ -0,0 +1,507 @@
...
+          className={cn(
+            'absolute inset-0 flex items-center justify-center bg-muted/20',
+            mode !== 'preview' && 'invisible pointer-events-none',
+          )}
+          aria-hidden={mode !== 'preview'}
+        >
+          <iframe
+            ref={iframeRef}
+            title="Slides preview"
+            sandbox="allow-scripts allow-same-origin allow-downloads"
```
Comment (Dr0p42) [inline libs/naas-abi/naas_abi/apps/nexus/apps/web/src/app/workspace/[workspaceId]/slides/[slug]/page.tsx]: [P0] Do not execute user/agent-editable deck HTML with both `allow-scripts` and `allow-same-origin`. For `srcDoc`, this lets arbitrary deck scripts run in the Nexus origin, access the parent frame/browser storage, and make authenticated same-origin requests—a stored XSS. Please render previews on an isolated origin or remove `allow-same-origin` and bridge only the required export operation through a constrained `postMessage`/server-side API.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
```diff
@@ -479,26 +715,61 @@ async def list_org_members(
...
 
     if member is None:
-        raise HTTPException(status_code=404, detail="User not found with this email")
-    return OrganizationMember(**member.__dict__)
+        raise HTTPException(status_code=500, detail="Failed to create or find user for invite")
+
+    workspace_member_id: str | None = None
+    if invite.workspace_id:
+        try:
+            ws_member = await workspace_service.invite_workspace_member(
```
Dr0p42, line 805: [P0] Validate that `invite.workspace_id` belongs to `org_id` before adding the member. The only authorization above is the caller's role in the organization, while `WorkspaceService.invite_workspace_member` accepts any workspace ID and performs no caller/org check. An admin of org A can therefore attach this user to an arbitrary workspace in org B if they know its ID. Resolve the workspace from `service.list_all_workspaces(org_id)` (or independently require workspace-admin access) and reject mismatches. The same optional-workspace flow in `nexus_admin_tools.py` needs the same guard.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
```diff
@@ -433,35 +487,217 @@ async def upload_org_logo_rectangle(
...
+        invalid = sorted({f for f in features if f not in known})
+        if invalid:
+            raise HTTPException(
+                status_code=400,
+                detail=f"Unknown features for {role_name}: {', '.join(invalid)}",
+            )
+        cleaned[role_name] = [f for f in features if f in enabled]
+
+    current = settings.feature_flags
+    settings.feature_flags = FeatureFlagsConfig(
```
Dr0p42: [P1] This org-scoped endpoint should not mutate deployment-global feature policy. Any owner/admin of any organization can reach this line and replace `settings.feature_flags`, immediately changing access for every organization/workspace handled by this process (and producing inconsistent policy across multiple workers). Store overrides per organization/workspace, or make a clearly deployment-scoped operation superadmin-only. The update should also preserve or require all baseline roles rather than silently dropping omitted roles.

libs/naas-abi-cli/naas_abi_cli/cli/user.py
```diff
@@ -0,0 +1,424 @@
...
+) -> None:
+    """Invite a user (create-on-invite) to an org and/or workspace.
+
+    Creates the Nexus account when missing and emails OTP / magic-link sign-in.
+    Prefer --org (optionally with --workspace). Workspace-only invite remains supported.
+    """
+    if not organization_id and not workspace_id:
+        raise click.ClickException("Provide --org and/or --workspace.")
+
+    client = build_client(api_url=api_url, token=token, email=email, password=password)
```
Dr0p42: [P1] Defer `build_client` until a non-dry-run branch. As written, `abi user invite ... --dry-run` requires credentials before it can print the request, so the newly added `test_user_invite_dry_run` fails with `Nexus auth required` on this PR head. I reproduced this with the focused CLI test suite.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/core/config.py
```diff
@@ -388,17 +392,22 @@ class Settings(BaseSettings):
     magic_link_expire_minutes: int = 15
     magic_link_max_active: int = 5
     magic_link_path: str = "/auth/magic-link"
+    otp_code_length: int = 6
```
Dr0p42: [P2] Validate these OTP settings at configuration load. `generate_otp_code()` falls back to 6 digits when `otp_code_length` is outside 4–10, but `verify_otp()` still requires the unvalidated configured length, making every generated code unverifiable for such a configuration. Using constrained fields (for example `Field(default=6, ge=4, le=10)` and `otp_max_attempts >= 1`) keeps generation and verification consistent.

Dr0p42 (changes requested):
I reviewed the current tip as a unit. Requesting changes for the tenant-boundary and stored-script-execution issues called out inline, plus the reproducible CLI dry-run regression.

Local checks on this SHA: frontend `tsc --noEmit` passed; 90 focused Python tests passed and 1 newly added test failed (`test_user_invite_dry_run`); changed Python files also have one Ruff failure (`UP017` in the Slides adapter at line 733).

jravenel:
## Tip note (Slides template gallery)

**Product flow:** File → New Presentation → Templates gallery (show all) → pick seed → title → Create → editor Preview + Abi customize (same open-deck loop).

**Shipped (4 templates):**
- `default-v1` Classic Board
- `pitch-dark-v1` Pitch Dark
- `minimal-light-v1` Minimal Light
- `executive-v1` Executive

**API:** `GET /api/slides/templates` discovers `*.html` + `catalog.json`. Create takes `template_id`.

**SHAs:** ABI `b83958e59` / Zen `07920f4` on `zen-next-abi-20260730`. Live on zen.naas.ai after IAP rebuild (`nexus-web` + `abi`).

**Add a template:** drop `<id>.html` (Zen contract: `.deck` / `.slide`, menubar, `buildPptx`) into ABI + Zen `assets/slides/templates/`, register in `catalog.json`. MIT NOTICE for Frontend Slides aesthetic inspiration only (not vendored).

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
jravenel: Addressed in 3b3e8c210.

- New decks are namespaced under `slides/<workspace_id>/<slug>` (branch + path) with `workspace_id` persisted in `project.json`.
- List/get/put/history/tree/runtime all resolve ownership before acting; cross-workspace access returns 404.
- Legacy `slides/<slug>` remains only when owned (or unscoped and claimed on open/save).
- Removed repo-wide `add_collaborator(..., permission="write")` as per-deck authorization; Nexus writes via the service path and only calls `ensure_user` for authorship.

libs/naas-abi/naas_abi/apps/nexus/apps/web/src/app/workspace/[workspaceId]/slides/[slug]/page.tsx
Comment (jravenel) [reply libs/naas-abi/naas_abi/apps/nexus/apps/web/src/app/workspace/[workspaceId]/slides/[slug]/page.tsx]: Addressed in 9910fd9e5.

Preview iframe sandbox is now `allow-scripts allow-downloads` (no `allow-same-origin`). Height metrics and PPTX export go through a constrained `postMessage` bridge injected into `srcDoc` (`zen-slides-preview`), so deck scripts cannot touch Nexus storage or make credentialed same-origin requests.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
jravenel: Addressed in e4fe2493a.

Org invite now rejects `workspace_id` values that are not in `list_all_workspaces(org_id)` before calling `invite_workspace_member`. The same guard is in `invite_organization_member` in `nexus_admin_tools.py`.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
jravenel: Addressed in 1db5a9920.

PUT no longer replaces deployment-global `settings.feature_flags.role_baseline`. It writes `organization_overrides[org_id]` only, requires all baseline roles (`owner`/`admin`/`member`/`viewer`), and `build_feature_flags` resolves the org overlay when building workspace feature flags. Deployment baseline stays intact for other orgs.

libs/naas-abi-cli/naas_abi_cli/cli/user.py
jravenel: Addressed in 4ad4c2dd1.

`abi user invite --dry-run` no longer calls `build_client` before printing the request. Auth runs only on the non-dry-run POST path, so `test_user_invite_dry_run` passes without credentials.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/core/config.py
jravenel: Addressed in fad444a40.

`otp_code_length` is now `Field(default=6, ge=4, le=10)` and `otp_max_attempts` is `Field(default=5, ge=1)`, so invalid configs fail at load and generation/verification stay consistent.

.deploy/docker/coder_prototype/template/abi_sidecar.py
```diff
@@ -0,0 +1,167 @@
...
+    def _send(self, code: int, payload: dict) -> None:
+        body = json.dumps(payload).encode("utf-8")
+        self.send_response(code)
+        self.send_header("Content-Type", "application/json")
+        self.send_header("Content-Length", str(len(body)))
+        self.end_headers()
+        self.wfile.write(body)
+
+    def _authed(self) -> bool:
+        if not SECRET:
```
Dr0p42: [P0] Fail-open auth on an endpoint that executes arbitrary shell commands.

`SECRET = os.environ.get("ABI_SIDECAR_SECRET", "")` defaults to empty, and this returns `True` for every request when it is. `TOOLS` includes `run_terminal`, which is `subprocess.run(command, shell=True)`, and line 167 binds `0.0.0.0:8378`. An empty secret therefore turns the sidecar into an unauthenticated remote shell reachable by anything on the docker network.

Nexus's own path is fine: `_prepare` always sets `sidecar_secret: ws_secret` from `secrets.token_hex(16)`, and `ensure_runtime` deletes rows with a missing secret to force reprovision. The exposure is the template default — both `template/main.tf` and `slides_template/main.tf` declare `sidecar_secret` with `default = ""` and export it straight into `ABI_SIDECAR_SECRET`, so any workspace created from these templates outside the Nexus ensure path (Coder UI, `coder create`, a template applied with defaults) starts wide open.

A missing secret should be a hard startup failure, not a fallback:

```python
SECRET = os.environ.get("ABI_SIDECAR_SECRET", "")
if not SECRET:
    raise SystemExit("ABI_SIDECAR_SECRET is required")
```

and drop the `if not SECRET: return True` branch. If a keyless dev mode is genuinely wanted, gate it on an explicit `ABI_SIDECAR_INSECURE=1` so it can never be reached by omission.

Minor, same file: `GET /health` answers before any auth check and returns `{"ok": true, "root": ROOT}`, which discloses the filesystem path and makes an open sidecar trivially discoverable by sweeping the network.

libs/naas-abi/naas_abi/agents/tools/nexus_admin_tools.py
```diff
@@ -0,0 +1,785 @@
...
+        Only provided (non-empty) fields are changed: name, email, company, role, bio.
+        Admins cannot edit another user's profile via API; invite/role tools manage access.
+        """
+        user_id = _require_user_id()
+        if isinstance(user_id, dict):
+            return user_id
+
+        payload = {
+            "name": name.strip() or None,
+            "email": email.strip() or None,
```
Dr0p42: [P0] This lets the LLM change the signed-in user's email address, which is an account-takeover primitive.

`AuthService.update_profile` (`services/auth/service.py:231`) checks only that the new address is not already taken — no re-authentication, no confirmation to the old address, no verification of the new one. Sign-in is keyed on email via `verify_otp` / `request_magic_link`, so once the address changes, the next code goes to whoever now owns it.

That risk is acceptable behind `PATCH /auth/me`, where the user typed it. It is not acceptable on a tool the model can call on its own: AbiAgent reads deck HTML, KG content, and workspace files, so a single injected instruction turns into a full account takeover with no user action.

Suggest dropping `email` from the tool signature and leaving it to the authenticated settings UI:

```python
def update_my_profile(name: str = "", company: str = "", role: str = "", bio: str = "") -> Any:
```

If email really must be agent-editable, it needs a confirm-to-new-address round trip before the change lands.

Same file, applies to all twelve tools: every one ends `except Exception as exc: return {"error": str(exc)}`. Because that string is returned to the model and rendered to the user, a SQLAlchemy/asyncpg connection failure puts the full DSN — password included — into the chat, and this PR also adds transcript export to file (`chat-transcript-export.ts`). Log the exception and return a fixed message instead.

libs/naas-abi/naas_abi/agents/tools/nexus_admin_tools.py
```diff
@@ -0,0 +1,785 @@
...
+        logger.debug("ABIModule nexus database_url unavailable: %s", exc)
+
+    import os
+
+    pg_host = (os.getenv("POSTGRES_HOST") or "").strip()
+    pg_user = (os.getenv("POSTGRES_USER") or "").strip()
+    pg_password = os.getenv("POSTGRES_PASSWORD") or ""
+    if pg_host and pg_user and pg_password and pg_host not in {"localhost", "127.0.0.1"}:
+        candidates.append(
+            f"postgresql+asyncpg://{pg_user}:{pg_password}@{pg_host}:5432/nexus"
```
Dr0p42: [P1] The DSN this builds decides which database the admin mutations write to, and it is assembled by string-sniffing.

Three concrete problems:

1. **The password is not URL-encoded.** `POSTGRES_PASSWORD` containing `@`, `/`, `:`, or `#` produces a malformed DSN or silently points at a different host. Needs `urllib.parse.quote_plus`.
2. **Port and database name are hardcoded** to `:5432/nexus`, ignoring `POSTGRES_PORT` and `POSTGRES_DB`.
3. **The `localhost` filter is inverted for single-host deployments.** The selection loop below skips every candidate containing `localhost`/`127.0.0.1`, so on a dev machine or a single-host install where Postgres genuinely is local, all candidates are rejected and it falls through to `candidates[0]` — which is not necessarily the URL the API process itself uses. Admin writes can land in a different database than the one serving the app, and there is a hardcoded `nexus:nexus@localhost:5432/nexus` at the end of the fallback chain.

Rather than re-deriving this, take the URL from one injected source of truth (the engine config) and fail loudly when it is absent — a wrong answer here is worse than an error.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
```diff
@@ -0,0 +1,1834 @@
...
+        if owner is not None and owner != workspace_id:
+            return None
+        return {**ns, "legacy": "0"}
+    legacy = _paths_for(workspace_id, slug, legacy=True)
+    if legacy["branch"] in names:
+        meta = _load_project_meta(
+            sc, repo_id=repo_id, path=legacy["project_path"], ref=legacy["branch"]
+        )
+        owner = _meta_workspace_id(meta)
+        if owner is not None and owner != workspace_id:
```
Dr0p42: [P1] The legacy fallback still fails open, which leaves a narrower version of the tenant boundary the namespacing fix closed.

Access is denied only when `owner is not None and owner != workspace_id`. When `owner is None` the legacy paths are returned and access is granted — and `owner` is `None` in two distinct cases:

- `project.json` is absent or has no `workspace_id` (every deck predating `3b3e8c210`). `_claim_workspace_in_meta` then stamps whichever workspace touched it first, so a legacy deck belongs to the first workspace to guess its slug rather than to its actual owner. Slugs come from `_slugify(title)`, so they are guessable, and `_repo_id()` is one shared repo across all tenants.
- `_load_project_meta` (line 213) swallows `SourceControlError` and returns `{}`. A transient Forgejo error while reading an **owned** legacy deck's `project.json` makes it look unowned for the duration of the blip.

The second case is the one I would fix first — an unreadable ownership record should deny, never grant. Distinguishing "read succeeded, no `workspace_id`" from "read failed" would be enough:

```python
def _load_project_meta(...) -> dict | None:   # None == unreadable
    try:
        return _parse_meta(sc.get_file(...).text)
    except SourceControlError:
        return None
```

and treat `None` as not-owned-by-caller. For the first case, a one-off backfill that stamps `workspace_id` from the existing deck owner would be safer than first-touch claiming; if the claim-on-open path stays, it is worth logging each claim so an unexpected cross-tenant claim is at least visible after the fact.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
```diff
@@ -0,0 +1,1834 @@
...
+router = APIRouter(dependencies=[Depends(get_current_user_required)])
+
+_SLUG_RE = re.compile(r"^[a-z0-9]+(?:-[a-z0-9]+)*$")
+_BRANCH_PREFIX = "slides/"
+_DEFAULT_TEMPLATE = "minimal-light-v1"
+_SIDECAR_PORT = 8378
+_SLIDES_TEMPLATE_NAMES = ("abi-slides", "abi-code-server")
+# Cold start: agent connect + startup_script before :8378 listens. Ensure must
+# wait; a single probe races "running" phase and falsely marks degraded.
+_SIDECAR_WAIT_ATTEMPTS = 40
```
Dr0p42: [P2] 40 attempts x 2.0s is an 80-second blocking wait held on a threadpool worker.

`ensure_runtime` reaches this via `await run_in_threadpool(_wait_for_sidecar, ws_base, ws_secret)`, and `_wait_for_sidecar` uses blocking `urlopen` + `time.sleep`. anyio's default thread limiter is 40 tokens, so roughly 40 concurrent deck opens pin the entire pool for 80s — starving every other `run_in_threadpool` call in the process, which in this adapter is all the Forgejo I/O. That is a self-inflicted outage at a very modest number of simultaneous users.

80s also exceeds most proxy and browser idle timeouts, so the caller often will not see the result it waited for.

Better shape: return `provisioning` immediately and let the client poll `GET /projects/{slug}/runtime`, which keeps the pool free and gives the UI something to show. If the synchronous wait stays for now, an async client with `asyncio.sleep` avoids consuming a thread at all, and the ceiling wants to be well under the proxy timeout.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
```diff
@@ -433,35 +490,242 @@ async def upload_org_logo_rectangle(
...
+    else:
+        workspaces = await service.list_workspaces(org_id=org_id, user_id=current_user.id)
     return [
+        _workspace_payload(ws, org_logo_url=org_logo_url, org_logo_rect_url=org_logo_rect_url)
+        for ws in workspaces
+    ]
+
+
+@router.get("/{org_id}/member-workspaces")
+async def list_org_member_workspaces(
```
Dr0p42, line 543: [P1] This route hands the full org workspace catalog and membership map to any org member, undoing the restriction added 40 lines above.

`list_org_workspaces` was deliberately changed in this PR so that non-admin members see only workspaces they belong to. This route calls only `require_org_access` and then returns `list_all_workspaces(org_id)` **plus** `list_workspace_memberships(org_id)` — every workspace in the org and who is in each one, to any member including `viewer`.

Since it exists to power the admin Workspaces column, it should carry the same gate as the sibling PUT:

```python
role = await require_org_access(current_user.id, org_id, service)
if role not in ("owner", "admin"):
    raise HTTPException(status_code=403, detail="Only admins can list member workspaces")
```

`GET /{org_id}/roles/features` below has the same shape — readable by any member while the PUT is admin-only. Lower impact since it is policy rather than people, but worth aligning.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
```diff
@@ -479,26 +743,67 @@ async def list_org_members(
...
+    workspace_service: WorkspaceService = Depends(get_workspace_service_for_org_invite),
+    auth_service: AuthService = Depends(get_auth_service),
+    email_service: EmailService | None = Depends(_get_email_service),
 ) -> OrganizationMember:
     role = await require_org_access(current_user.id, org_id, service)
     if role not in ("owner", "admin"):
         raise HTTPException(status_code=403, detail="Only admins can invite members")
 
+    email = invite.email.lower().strip()
+    _user, user_created = await auth_service.ensure_user_for_invite(email, name=invite.name)
```
Dr0p42, line 788: [P2] The account is created and committed before the request is validated, and nothing rate-limits this endpoint.

`ensure_user_for_invite` calls `await self.adapter.commit()` internally, so by the time the `workspace_id`-belongs-to-org check added in `e4fe2493a` runs (line ~779) the user and their personal workspace already exist. A caller who posts a bad `workspace_id` gets a 400 and leaves behind an orphaned account. Reordering so all validation precedes the first mutation fixes it, and it would be worth having the whole handler share one transaction rather than committing in three places.

Separately: this creates real, sign-in-able accounts for arbitrary addresses and emails each one a live OTP, with no rate limit and explicitly bypassing `magic_link_allow_signup`. Any org admin can mint unbounded accounts and send unsolicited sign-in mail to addresses they do not control — worth the same `check_rate_limit` treatment the auth routes get, both here and on the workspace invite route.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
```diff
@@ -433,35 +490,242 @@ async def upload_org_logo_rectangle(
...
+            )
+        cleaned[role_name] = [f for f in features if f in enabled]
+
+    current = settings.feature_flags
+    org_overrides = {
+        key: {r: list(feats) for r, feats in overlay.items()}
+        for key, overlay in current.organization_overrides.items()
+    }
+    org_overrides[org_id] = cleaned
+    settings.feature_flags = FeatureFlagsConfig(
```
Dr0p42: [P2] Follow-up on the earlier thread — the blast radius is correctly scoped now, but the storage is still a process-global.

`1db5a9920` fixed the real problem: this no longer replaces the deployment baseline for every org. Three things remain, all from `settings` being in-process mutable state written from a request handler:

1. **Not persisted** — every override is lost on restart, while the UI presents it as a saved setting.
2. **Not shared across workers** — with more than one uvicorn worker or replica, only the process that served the PUT has the override, so a user's feature flags depend on which worker answers. That is the failure mode most likely to be reported as a bug.
3. **Read-modify-write race** — two concurrent PUTs for different orgs both copy `current.organization_overrides`, and the second to assign drops the first's change.

The `note` in the GET response is honest about this, which I appreciate, but it is not visible to anyone using the UI. This wants a table.

Minor, just above at line ~522: `cleaned[role_name] = [f for f in features if f in enabled]` silently drops any feature not in `enabled_features`. An admin ticks a box, gets 200, and the box comes back empty with no explanation. Rejecting with 400, or returning the dropped keys, would save a support round-trip.

libs/naas-abi/naas_abi/apps/nexus/apps/web/src/components/slides/slides-preview-frame.tsx
```diff
@@ -0,0 +1,172 @@
...
+    if (!host || typeof ResizeObserver === 'undefined') return;
+    const ro = new ResizeObserver(() => measureHost());
+    ro.observe(host);
+    return () => ro.disconnect();
+  }, [measureHost]);
+
+  useEffect(() => {
+    const onMessage = (event: MessageEvent) => {
+      if (event.source !== iframeRef.current?.contentWindow) return;
+      if (!isSlidesPreviewMessage(event.data)) return;
```
Dr0p42, line 82: [P1] The handler validates the payload shape but not the sender, so preview frames cross-talk.

`isSlidesPreviewMessage` only checks `event.data`. Every mounted `SlidesPreviewFrame` listens on the same global `window` message channel, so with more than one on screen — and this PR adds a dual compare pane — a `metrics` message from frame A calls `setDocHeight` on every instance, and a single `export-pptx-result` resolves or rejects the waiters of all of them. That is a real correctness bug before it is a security one.

The origin cannot be checked here (the sandbox has no `allow-same-origin`, so `event.origin` is `"null"`), but the source can:

```js
if (event.source !== iframeRef.current?.contentWindow) return;
if (!isSlidesPreviewMessage(event.data)) return;
```

That scopes each listener to its own frame and incidentally removes the spoofing channel from any other frame on the page.

Dr0p42 (changes requested):
Re-reviewed at tip after the six previous threads were addressed. Confirmed fixed on this branch:

| Prior thread | Verified |
|---|---|
| Slides tenant boundary (`3b3e8c210`) | Namespaced `slides/<workspace_id>/<slug>`, ownership resolved on every route, repo-wide `add_collaborator(write)` gone — **but see the legacy fallback thread below** |
| Preview iframe stored XSS (`9910fd9e5`) | `sandbox="allow-scripts allow-downloads"`, no `allow-same-origin`, `postMessage` bridge — **one gap noted below** |
| Org invite `workspace_id` scoping (`e4fe2493a`) | Enforced in both the FastAPI adapter and `nexus_admin_tools.py` |
| Role features global mutation (`1db5a9920`) | No longer replaces the deployment baseline; per-org overlay + all-roles requirement |
| CLI `--dry-run` auth (`4ad4c2dd1`) | `build_client` deferred past the dry-run branch |
| OTP config constraints (`fad444a40`) | `Field(default=6, ge=4, le=10)` / `ge=1` |

Requesting changes again for two new issues I consider blocking, plus a handful of correctness items. All nine are inline. The two blockers:

**1. The workspace sidecar treats a missing secret as "auth disabled" on an endpoint that runs arbitrary shell commands.** `_authed()` returns `True` when `ABI_SIDECAR_SECRET` is empty, `run_terminal` is `subprocess.run(..., shell=True)`, and the server binds `0.0.0.0:8378`. Nexus's own provisioning path always passes a `secrets.token_hex(16)` secret and deletes rows with a missing secret to force reprovision, so the happy path is covered — but both `main.tf` files declare `sidecar_secret` with `default = ""`, so any workspace created from these templates outside that path comes up as an unauthenticated remote shell on the docker network. This should be a hard startup failure, not a fallback.

**2. `update_my_profile` lets the LLM change the signed-in user's email address.** `AuthService.update_profile` performs no re-auth and sends no confirmation to the old or new address, and sign-in is keyed on email — so changing it hands the account to whoever controls the new address. AbiAgent also reads untrusted content (deck HTML, KG, files), which makes injection → takeover a single hop. Dropping `email` from the tool signature closes it.

### Everything else, by priority

- **P1** — Legacy slides paths still fail open when `project.json` is unreadable or lacks `workspace_id`; first toucher claims the deck, and slugs are `_slugify(title)`, so guessable.
- **P1** — `_resolve_database_url()` picks the write target by substring-sniffing for `localhost`, hardcodes `:5432/nexus`, and does not URL-encode the password. Compounded by `return {"error": str(exc)}` on every tool, which puts the DSN (password included) into the chat transcript — and this PR also adds transcript export to file.
- **P1** — `GET /{org_id}/member-workspaces` returns the full org workspace catalog and membership map to any org member, undoing the restriction this same PR adds to `list_org_workspaces` two routes above.
- **P1** — The preview `message` handler validates payload shape but not `event.source`, so with two preview frames mounted (this PR adds a dual compare pane) metrics from one resize both and one export result resolves every instance's waiters.
- **P2** — `_SIDECAR_WAIT_ATTEMPTS = 40` × `2.0s` = 80s held on an anyio threadpool worker (default pool: 40), so ~40 concurrent deck opens starve every other threadpool endpoint.
- **P2** — Invite flow is not atomic and has no rate limit: `ensure_user_for_invite` commits before validation, so a later 400 leaves an orphaned user + personal workspace.
- **P2** — Per-org role overlays live in a process-global mutated from a request handler: lost on restart, inconsistent across workers, and concurrent PUTs for different orgs lose updates. The `note` field is honest, but the UI reads as persisted.

Two more that do not have a natural anchor in the diff:

- **OTP codes are hashed with unsalted SHA-256** (`hash_token` in `services/refresh_token.py`). That is fine for a 32-byte magic-link token; for a 6-digit code the keyspace is 10⁶ and a rainbow table inverts it instantly. Anyone with read access to `magic_link_tokens` — a backup, a replica, a read-only injection — recovers live codes and signs in as any user with a pending challenge. An HMAC with a server-side key would fix it without changing the schema.
- **The OTP is in the email Subject header and logged in plaintext.** `magic_link_email_subject_template` embeds `{otp_code}`; subjects surface on lock screens and are retained by relays and SIEMs more readily than bodies. Separately, when `email_service is None` both `_send_magic_link_email` and `invites/sign_in_email.py` log the live code at INFO — worth gating behind an explicit dev flag so a silent email-transport failure in prod does not write credentials to the app log.

CI is pending on this tip at the time of review, so I have not re-run the suites; the findings above are from reading the code at `zen-next-abi-20260730`.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
```diff
@@ -0,0 +1,1834 @@
...
+        return {**ns, "legacy": "0"}
+    legacy = _paths_for(workspace_id, slug, legacy=True)
+    if legacy["branch"] in names:
+        meta = _load_project_meta(
+            sc, repo_id=repo_id, path=legacy["project_path"], ref=legacy["branch"]
+        )
+        owner = _meta_workspace_id(meta)
+        if owner is not None and owner != workspace_id:
+            return None
+        return {**legacy, "legacy": "1"}
```
Dr0p42, line 257: [P0] Fail closed for unowned legacy decks instead of treating missing ownership as authorization. `_load_project_meta` converts a missing/malformed metadata file or any `SourceControlError` into `{}`, and this branch then returns the legacy paths; `get_project`/`get_deck`/`put_deck` immediately claim them for whichever workspace guessed the slug. That still permits cross-workspace read/write takeover and can also strip isolation during a transient metadata read failure. Require a verified matching owner here, and move any legacy claim/migration into an explicit privileged flow; the agent-side resolver should follow the same rule.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
Dr0p42: [P1] Avoid holding each runtime request open for this full poll window. A probe can itself block for 2 seconds, so 40 probes plus 2-second sleeps can take roughly 80–160 seconds; `create_project` waits here synchronously, and the client then retries the runtime endpoint up to six times. This can exceed normal proxy/request timeouts and consume the threadpool for minutes per cold deck. Return the provisioning state promptly and let the existing client polling observe readiness, or bound the server-side wait to a short request-safe interval/background task.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
```diff
@@ -0,0 +1,1834 @@
...
+            external_id=current_user.id,
+            email=author_email,
+            username=username,
+        )
+        existing = {b.name for b in sc.list_branches(repo_id=repo_id)}
+        if branch in existing:
+            raise BranchNameConflictError(f"Slides project '{slug}' already exists")
+        # Block silent overwrite of a legacy same-slug deck owned by anyone.
+        legacy_branch = _legacy_branch_for(slug)
+        if legacy_branch in existing:
```
Dr0p42: [P1] Do not let a legacy branch reserve this slug across every workspace. The new branch is namespaced as `slides/<workspace_id>/<slug>`, so creating it cannot overwrite `slides/<slug>`; this check instead prevents all other tenants from ever using a slug that exists in legacy storage, preserving the cross-tenant slug-collision problem the namespace was intended to remove. Only reject the current workspace's namespaced branch; handle legacy migration/ownership separately.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/core/config.py
```diff
@@ -222,12 +225,18 @@ class FeatureFlagsConfig(BaseModel):
                 "ontology",
                 "graph",
                 "settings",
+                "slides",
             ],
-            "member": ["maps", "chat", "files", "skills"],
-            "viewer": ["maps", "chat", "files", "skills"],
+            "member": ["maps", "chat", "files", "skills", "slides"],
```
Dr0p42, line 230: [P1] Keep the colocated feature-flag test suite consistent with this default. At the current tip, `uv run --frozen pytest .../core/feature_flags_test.py` fails because `test_member_role_defaults_to_chat_and_files_only` still expects the old mapping without `slides` (74 other focused tests pass). Update the expected member/viewer defaults and test name if enabling Slides is intentional, or revert this default; the current GitHub workflow does not exercise this colocated test.

Dr0p42 (changes requested):
Re-reviewed the 18 commits added since the prior local review tip (`1247d7baa`) through current head `26b1374fd`.

The earlier CLI dry-run, OTP validation, org-invite workspace validation, and removal of iframe `allow-same-origin` are present. I am still requesting changes for the remaining Slides tenant-boundary issue and the blocking cold-start behavior called out inline.

Local verification: 74/75 focused Python tests passed; `feature_flags_test.py::TestBuildFeatureFlags::test_member_role_defaults_to_chat_and_files_only` fails because the new `slides` default was not reflected in the assertion. Frontend `tsc --noEmit` passed and the three Vitest-native focused suites passed. GitHub's current PR Checks workflow is green, but it does not catch that colocated Python test.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
Dr0p42: This remains unresolved. The mapping is now keyed per organization, but `PUT` still assigns it only to this process's in-memory `settings.feature_flags`. With multiple API workers, the next request can hit a worker that never saw the override; every restart also discards it. Persist the org override through an organization-owned port/store (and load it when building flags), or keep this as a deployment-config-only operation instead of presenting the save as an organization update.

.deploy/docker/coder_prototype/template/abi_sidecar.py
jravenel: Addressed in aa2c8313a.

- Missing `ABI_SIDECAR_SECRET` is now a hard startup failure.
- Fail-open auth is gone; keyless mode requires explicit `ABI_SIDECAR_INSECURE=1`.
- `/health` requires the bearer token and no longer returns `ROOT`.
- Template parameters no longer default the secret to an empty string.

libs/naas-abi/naas_abi/agents/tools/nexus_admin_tools.py
jravenel: Addressed in def561590.

- `update_my_profile` no longer accepts `email`; email changes stay in the authenticated settings UI.
- All twelve tools now log the real exception server-side and return a fixed error string, so DSNs cannot reach the chat transcript.

libs/naas-abi/naas_abi/agents/tools/nexus_admin_tools.py
jravenel: Addressed in def561590.

`_resolve_database_url` now prefers `NEXUS_DATABASE_URL` / `DATABASE_URL`, then ABIModule config, then URL-encoded `POSTGRES_*` (including `POSTGRES_PORT` / `POSTGRES_DB`), then Settings. It fails loudly when none are available, and the localhost filter / hardcoded fallback DSN are gone.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
jravenel: Addressed in 68ddba568.

- `_load_project_meta` returns `None` on `SourceControlError` (deny) vs `{}` on a successful empty/malformed read.
- Legacy paths require a verified matching `workspace_id`; unowned decks are not accessible and first-touch claim-on-open/save is removed.
- Agent `slides_tools` resolver follows the same fail-closed rule.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
jravenel: Addressed in 68ddba568.

Server-side sidecar wait is now 2 x 0.5s. `ensure_runtime` returns promptly with `sidecar_ready=false` when still cold, so the client poll path can observe readiness without pinning the threadpool for ~80s.

jravenel: Addressed in 68ddba568 (same change as the related wait thread).

Probe window is request-safe (2 x 0.5s). Create/ensure return provisioning state quickly; existing client polling covers readiness.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
jravenel: Addressed in 11e7f75bf.

`GET /{org_id}/member-workspaces` and `GET /{org_id}/roles/features` now require org owner/admin, matching the mutating routes.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
jravenel: Addressed in 11e7f75bf (and the matching admin-tool path in def561590).

- `workspace_id` is validated against `list_all_workspaces(org_id)` before `ensure_user_for_invite`.
- Org and workspace invite routes use `check_rate_limit` (caller + email), same pattern as auth.
- Full single-transaction invite rewrite is still a follow-up; this removes the orphan-on-bad-workspace_id path.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
jravenel: Partial follow-up in 11e7f75bf.

- Disabled/unknown features now return 400 instead of silently dropping.
- GET `note` now states clearly that overlays are in-process only (lost on restart / not shared across workers).

Durable `organization_overrides` storage still needs a table (or equivalent). Leaving this thread open for that follow-up.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
jravenel: Follow-up: blast-radius fix from 1db5a9920 remains. 11e7f75bf adds 400s for disabled features and a clearer non-durable note. Persisting org overlays to a table is still open (see the later thread).

libs/naas-abi/naas_abi/apps/nexus/apps/web/src/components/slides/slides-preview-frame.tsx
jravenel: Already present since 9910fd9e5 (line 81):

```ts
if (event.source !== iframeRef.current?.contentWindow) return;
if (!isSlidesPreviewMessage(event.data)) return;
```

Each `SlidesPreviewFrame` instance only handles messages from its own iframe. No further code change in this pass.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
jravenel: Addressed in 68ddba568.

Legacy access now requires a verified matching owner. Unreadable metadata denies. Claim-on-open/save is removed; agent resolver matches the API rule.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/slides/adapters/primary/slides__primary_adapter__FastAPI.py
jravenel: Addressed in 68ddba568.

`create_project` only conflicts on the namespaced `slides/<workspace_id>/<slug>` branch. A legacy `slides/<slug>` branch no longer blocks other tenants from creating the same slug under their workspace namespace.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/core/config.py
jravenel: Addressed in 923281d81.

Colocated `feature_flags_test` now expects `slides: True` for member defaults (`test_member_role_defaults_include_slides`). Focused suite passes.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
jravenel: Addressed in `56f3c5817`.

- New table `organization_role_features` (migration `0040`) stores the per-org `role_baseline` JSON.
- `PUT /organizations/{org_id}/roles/features` upserts that row (admin/owner only) and no longer mutates in-process `settings.feature_flags`.
- `GET` and workspace `feature_flags` resolve the durable overlay from the DB, so it is shared across workers and survives restarts.
- UI only shows Saved when the response `persistence` is `database`.

Live on zen.naas.ai after IAP rebuild (`abi` + `nexus-web` + `caddy`). Zen pin: `0bb1f81`.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
jravenel: Addressed in `56f3c5817` (same change as the follow-up on the earlier P1 thread).

Overlays are now in `organization_role_features`, so they persist across restarts, are shared across API workers, and avoid the in-process read-modify-write race on `settings.feature_flags.organization_overrides`. Disabled/unknown features are rejected with 400/422 instead of being silently dropped.

libs/naas-abi/naas_abi/apps/nexus/apps/web/src/app/workspace/[workspaceId]/maps/lib/datasets.ts
```diff
@@ -199,6 +210,17 @@ export const MAPS_DATASETS: MapsDataset[] = [
     icon: 'Laptop',
     order: 0,
   },
+  // Zen tip product overlay (not for abi main). Private stays presence-only;
+  // Custom hosts named intelligence layers such as Ontologist, North America.
+  {
+    id: 'ontologist-north-america',
```
Comment (Dr0p42) [inline libs/naas-abi/naas_abi/apps/nexus/apps/web/src/app/workspace/[workspaceId]/maps/lib/datasets.ts:216]: [P1] Keep this Zen-only layer out of upstream ABI. The comment immediately above says it is “not for abi main”, and the Nexus `AGENTS.md` contract says Custom remains empty upstream while product-specific datasets live in the product overlay. Upstream also has no `/api/intelligence/ontologist-north-america` backend route, so this visible catalog item always ends at a 502 in a normal ABI deployment; its UI copy additionally hard-codes Zen, Sanax, and NaasAI. Please move the registry/component/proxy/middleware/pins to the Zen overlay, or first add a generic configured extension contract and register the layer only when that backend is available.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
```diff
@@ -479,26 +758,75 @@ async def list_org_members(
...
+            )
+        except WorkspaceMemberAlreadyExistsError:
+            ws_member = None
+        if ws_member is not None:
+            workspace_member_id = ws_member.id
+
+    sign_in_email_sent = False
+    challenge = await issue_invite_sign_in_challenge(auth_service, email)
+    if challenge is not None:
+        sign_in_email_sent = await send_invite_sign_in_email(
```
Dr0p42, line 818: [P2] This route (and the workspace invite route) bypasses the new send-failure cleanup in `ensure_user_and_send_invite_email`. If `EmailService.send` throws after the user and membership have committed, the request returns 500 while leaving an active challenge that was never delivered; retrying then returns “already a member”, so the admin cannot resend through the same flow. Please call the shared helper here or apply the same invalidation and make an existing-member invite idempotently resend the sign-in challenge.

Dr0p42 (changes requested):
Re-reviewed the current tip after the latest follow-up fixes. The previously reported Slides tenant isolation, sidecar authentication, OTP handling, and durable role-feature storage changes are present. I am requesting changes for the upstream-scope/runtime blocker called out inline; I also found an invite-email failure path that bypasses the cleanup added elsewhere.

Local verification on this SHA: all changed Python files pass Ruff; 197 focused Python tests passed. Frontend typecheck was not run because the locked `vis-data` tarball was unavailable from the configured package host.

Dr0p42:
## Review: `zen-next-abi-20260730` → `main`

Reviewed at head `1c0bd27c9` (merge-base `010cd1fda`): **180 files, +18,356 / −2,292, 71 commits.**

First, credit where it's due — for a change this size the security fundamentals are notably careful, and I want to be specific because it shaped how I triaged everything else:

- OTP codes come from `secrets.randbelow`, are stored as **HMAC-SHA256 keyed on `secret_key`** (not plaintext, not a bare hash of a 10⁶ keyspace), compared with `hmac.compare_digest`, and are single-use with expiry enforced.
- `nexus_admin_tools.py` gets the **credential model right**: no service token, identity comes from `agent_user_id` set at the stream boundary from `current_user.id`, and every mutating tool re-reads the caller's role from the DB. It fails closed when unauthenticated, all 12 call sites guard the union return, and `_tool_error` scrubs exceptions so DSNs never reach the model.
- All 9 slides endpoints sit behind `get_current_user_required` **and** a per-request `require_workspace_access`; slugs are regex-validated before any path construction; the preview iframe uses `sandbox="allow-scripts allow-downloads"` **without** `allow-same-origin`.
- The Coder sidecar has no default secret, aborts startup without one, and uses a constant-time bearer compare.

The findings below are what survived an adversarial verification pass — several candidate findings were refuted and dropped, and I re-verified every item here against the code myself.

---

## 🔴 Blockers

### 1. The new OTP endpoint has no effective brute-force protection in any deployment built from this repo

`/api/auth/otp/verify` calls `check_rate_limit`, so this looks covered — but it isn't, for two compounding reasons:

**(a) Rate limiting is off by default.** `config.py:429` disables it when `environment == "development"` **or** `nexus_env == "local"`:

```python
def model_post_init(self, __context: Any) -> None:
    if self.environment == "development" or self.nexus_env == "local":
        self.rate_limit_enabled = False
```

`environment` defaults to `"development"` (`config.py:357`), and `ENVIRONMENT` **is never set anywhere in this repository** — I grepped the entire PR tree. `docker-entrypoint.sh` sets `NEXUS_ENV=${NEXUS_ENV:-cloudflare}`, but the condition is `or`, so the `environment` default alone disables the limiter.

**(b) `otp_max_attempts` is not a real guess budget.** In `verify_otp` (`service.py:604`), a failed guess is charged to `active[0]` only, while the guess is tested against *every* active challenge (up to `magic_link_max_active=5`). Any unauthenticated `POST /magic-link/request` mints a fresh row with `otp_attempts=0` and resets the budget.

Net effect: unlimited guesses against a 6-digit code (10⁶) for any known email address → account takeover.

The auto-disable is **pre-existing** and not introduced here, but this PR is what makes it critical: a short numeric code is far more brute-forceable than the password login it previously guarded. Worth confirming what Zen actually sets `ENVIRONMENT` to in its runtime env — if it's unset there too, this is live.

**Suggested fix:** enforce OTP throttling independently of `rate_limit_enabled` (or make the limiter fail-closed in non-dev), and charge the failed attempt to *every* challenge the guess was tested against.

### 2. Slides tools trust a client-supplied `workspace_id` as the tenancy boundary

At `chat__primary_adapter__streaming.py:254`:

```python
if request.workspace_id is not None:
    agent_workspace_id.set(str(request.workspace_id))
```

`request.workspace_id` comes straight from the request body, and there is no `require_workspace_access` on this path. `slides_tools.py` then uses that contextvar as the tenancy key (`:70`) while only ever checking `if not agent_user_id.get()` — i.e. "is *someone* signed in", never "is this user a member of that workspace".

An authenticated user can POST to `/chat/stream` with another workspace's ID and have the slides tools read and write that workspace's decks. The UI would never do this; a direct API call would.

What makes this clearly an oversight rather than a design choice: **`nexus_admin_tools.py` in this same PR does it correctly**, calling `require_workspace_access` / `get_workspace_role` before acting. The slides path just didn't get the same treatment.

**Suggested fix:** call `require_workspace_access(current_user.id, request.workspace_id)` at the stream boundary before setting the contextvar — one check, and it covers every current and future context-scoped tool.

*(Checked and clear: `coder_workspace_base` / `coder_workspace_secret` are resolved server-side via `lookup_slides_sidecar`, not taken from client context — no SSRF or secret-injection there.)*

---

## 🟠 High

### 3. The OTP code never actually reaches the user's inbox

This PR updates the email templates in `config.py` to include `{otp_code}` — but **three** other copies still carry the old link-only text and were not updated:

| Location | Status |
|---|---|
| `config.local.yaml:82` | ❌ old template, **not touched by this PR** |
| `libs/naas-abi-cli/.../templates/project/config.local.yaml:66` | ❌ old template (every new project) |
| `libs/naas-abi/naas_abi/__init__.py:365` (`NexusConfig`) | ❌ old template |

```yaml
magic_link_email_text_template: "Use this link to sign in to {app_name}:\n\n{magic_link_url}\n\nThis link expires in {expire_minutes} minutes."
```

The consumer is `settings.magic_link_email_text_template.format_map(...)`. Because `format_map` silently ignores unused keys, there's no error — the email just renders **without the code**, while the new UI asks the user to type a 6-digit code they never received.

Since `config.local.yaml` is what production runs, the flagship feature of this PR appears broken end-to-end on the configs shipped in this tree. Worth a manual send-and-check before merge.

### 4. Slides uses an unpaginated branch listing as its project index

`_resolve_project_paths` (`:231`), `list_projects` (`:813`), `create_project` (`:898`) and `get_project` (`:1038`) all treat `sc.list_branches(...)` as a complete index of the shared monorepo. Forgejo paginates, so once the repo exceeds one page of branches, existing decks start returning 404 and `create_project` can collide with a slug it can't see. This degrades silently and gets worse with adoption.

### 5. Sidecar/Forgejo dual-write can silently lose committed deck edits

`put_deck` prefers stale sidecar content over Forgejo and then blind-writes it back, with no reconciliation or optimistic concurrency. The saver's own container stays in sync, but **every other workspace member's sidecar** can overwrite committed edits with a stale copy. No conflict is surfaced.

### 6. `_split_sections` silently truncates decks without a `<main>` wrapper

`slides_tools.py:566-574` — the no-`<main>` branch drops the tail after the last `</section>`. Because the result is written back and committed, this is **persisted, unrecoverable content loss** on any deck not matching the expected shape. The lossless handling already exists in `_attach_inter_section_markup`; this branch just doesn't use it.

### 7. Deck writes go to the bound sidecar even when the tool names a different slug

`_write_deck_via_sidecar` (`:210`) resolves the target purely from `coder_workspace_base`/`coder_workspace_secret`, which are bound to the *open* deck. If the model passes an explicit different `slug`, the write lands in the wrong deck's container. Gate the sidecar on `slug == slides_active_slug.get()` and fall back to Forgejo otherwise.

### 8. The auto-pushed Coder template installs a committed dev root CA into every workspace

`.deploy/docker/coder_prototype/template/main.tf:308`. Baking a static, source-committed CA into the trust store of every provisioned workspace means anyone with the repo holds the corresponding key. Please drop it, or generate per-deployment at bootstrap.

### 9. Scope: the "Ontologist, North America" map looks like it shouldn't be on `main`

`maps/lib/datasets.ts:213` is annotated in-tree with `// Zen tip product overlay (not for abi main).`, and it proxies an `/api/intelligence` endpoint that **does not exist in this repository** — so it cannot work here. It arrives with a route handler, component, pins fixture, middleware branch and tests.

Relatedly, **`/api/maps/gulf-strikes/route.ts` has no authentication at all** — no Bearer check, no 401 — while its sibling `ontologist-north-america` got both an explicit Bearer check *and* a dedicated middleware branch. It's a public-data proxy rather than a data leak, so it's not a blocker, but the asymmetry looks unintentional and it leaves an unauthenticated open proxy on an otherwise authenticated product.

---

## 🟡 Medium / Low

- **Agent invite tools skip rate limiting** that both REST invite endpoints enforce (`nexus_admin_tools.py:392`, `:617`) — an LLM-driven invite loop isn't throttled.
- **`provision()` adopt path drops `params`** (`CodingEnvironmentService.py:110`), restarting a workspace on the previous build's `repo_url`/branch/secrets.
- **Template push is one-shot** (`bootstrap-coding.sh:154`): gated on *total* template count, so once any template exists the bundled `main.tf` never updates. Gate on the template name instead.
- **Break-glass `--via postgres` skips email lowercasing** (`nexus_postgres.py:47`), producing an account that can never log in, since the auth service normalizes to lowercase everywhere else.
- **`abi user create` discards the generated password** if any post-register step fails (`user.py:166`).
- **Migration `0041` hard-deletes `agent_configs` rows** with no backup table and no row-count logging. The dedup is justified, but on a live DB I'd `CREATE TABLE ... AS SELECT` the losers first.
- **OTP enumeration oracle**: `ExpiredOtpError` vs `InvalidOtpError` distinguishes "account exists" from "doesn't" (`service.py:591`).
- **Failed OTP guesses invalidate the victim's magic link** (`service.py:606`) — unauthenticated sign-in denial.
- **Invite endpoints borrow the login brute-force budget**: the 6th invite in 5 minutes returns 429 (`organizations__primary_adapter__FastAPI.py:773`).

On the CLI's raw SQL — I specifically tried to break `esc()` and **could not**. Values stay inside quoted literals, Postgres defaults to `standard_conforming_strings=on`, psql doesn't expand `:vars` inside literals, and the SQL arrives on stdin rather than argv. It's fragile as a pattern and I'd still prefer parameter binding, but I'm not reporting it as a vulnerability.

---

## Process notes

- **The PR description is stale.** It pins tip `26b1374fd`, but head is `1c0bd27c9` — **21 commits ahead**. The stated Zen submodule pin therefore doesn't match what would merge. For a change explicitly framed as "single review vehicle", that's worth correcting so reviewers and the Zen pin agree.
- **Test coverage is thin on the riskiest new code:** slides API `174` test lines for `1,872` source; `nexus_admin_tools` `80` for `798`. The admin-tool tests are *well chosen* (unauthenticated rejection, bad slug, non-admin invite), but only one of ~12 tools has its authz-denial path covered — and that's the property most worth locking down.
- **A 1,872-line file named `..._primary_adapter__FastAPI.py` is carrying substantial business logic** (sidecar/Forgejo fallback, reconciliation, path resolution). Given the hexagonal conventions the rest of the repo follows, that logic probably belongs in a slides service with the adapter kept thin. Not a merge blocker, but it's where several bugs above are concentrated, which is itself evidence.

---

**Recommendation:** blockers 1 and 2 are small, targeted fixes (throttle OTP independently of the dev flag; add one `require_workspace_access` at the stream boundary) and I'd want both before this lands. Finding 3 is worth a manual verification since it likely means the headline feature doesn't work in production as configured. The slides data-loss items (5, 6, 7) are the ones I'd fix before real decks exist, because they're unrecoverable after the fact rather than merely broken.

<sub>Review assisted by Claude Code: 12 domain-scoped reviewers over the full diff, each finding adversarially verified by an independent skeptic (5 candidate findings refuted and dropped), then re-checked by hand against the source before posting. Pre-existing issues are labelled as such; where I couldn't reproduce a claim, I've said so.</sub>

libs/naas-abi/naas_abi/apps/nexus/apps/web/src/app/workspace/[workspaceId]/maps/lib/datasets.ts
Comment (Dr0p42) [reply libs/naas-abi/naas_abi/apps/nexus/apps/web/src/app/workspace/[workspaceId]/maps/lib/datasets.ts:216]: Fix proposed in #1131 (targets this branch, not `main`).

I took the second option you were offered here — a generic configured extension contract — because it removes the Zen specifics from upstream *and* leaves Zen a supported way to keep the layer without patching ABI.

- `NEXT_PUBLIC_MAPS_CUSTOM_DATASETS` (new `src/lib/maps-custom-datasets.ts`) holds a JSON array of layer descriptors. Custom is empty upstream again, and a layer only registers where an operator pointed it at a backend that exists.
- The Zen-specific proxy, component, pins fixture and middleware special-case are gone, replaced by `/api/maps/custom/[datasetId]`, `MapsCustomFeed`, and a `/api/maps/custom/` prefix gate.
- `endpoint` must be a Nexus API path — absolute and protocol-relative URLs are rejected at parse time, so the proxy can never forward the caller's Bearer token off-host. Ids must be route-safe and cannot shadow a built-in dataset.
- `AGENTS.md` documents the contract; `datasets.test.ts` asserts Custom is empty upstream.

Zen keeps the layer by setting the env var with `"endpoint": "/api/intelligence/ontologist-north-america"` — no ABI patch needed.

Leaving this thread open until #1131 lands.

libs/naas-abi/naas_abi/apps/nexus/apps/api/app/services/organizations/adapters/primary/organizations__primary_adapter__FastAPI.py
Dr0p42: Fix proposed in #1131 (targets this branch, not `main`).

- Extracted `issue_and_send_invite_sign_in` in `invites/sign_in_email.py`: it invalidates the challenge whether delivery raises *or* reports `False` (no transport configured). `ensure_user_and_send_invite_email` now wraps it, so there is a single guarded path.
- Both the org and workspace invite routes call it. So do the two `nexus_admin_tools.py` invite tools, which had the same inline pattern you flagged here.
- Re-inviting an existing member now resends idempotently instead of 400 (`status: "already_member"` on the workspace route); it still 400s when no membership row can be resolved.
- `stores/organization.ts` replaces rather than appends the returned membership, so a resend cannot duplicate the row in the UI cache.

Covered by 9 new tests (`sign_in_email_test.py`, `organizations__primary_adapter__FastAPI_invite_test.py`).

Leaving this thread open until #1131 lands.

Dr0p42:
## Open review threads: status

Went through the seven threads still unresolved at `1c0bd27c9`.

**Five were already fixed at head and just never resolved** — I verified each against the current tip and resolved them:

| Thread | Verified at `1c0bd27c9` |
|---|---|
| Org invite `workspace_id` scoping | Enforced in the FastAPI adapter (`list_all_workspaces` membership check) *and* `nexus_admin_tools.py` |
| Role features global mutation | `organization_role_features` + migration `0040`; per-org overlay, durable across workers/restarts |
| Role features process-global storage | Same change — resolved through the DB, not `settings` |
| CLI `--dry-run` auth | `build_client` deferred past the dry-run branch in both invite paths |
| OTP config constraints | `Field(default=6, ge=4, le=10)` / `otp_max_attempts` `ge=1` |

**The two from my 2026-07-31 review are genuinely open, and #1131 fixes both** (based on this branch and targeting it, so it lands inside this review vehicle):

1. **Zen-only Maps layer upstream** — took the "generic configured extension contract" option: `NEXT_PUBLIC_MAPS_CUSTOM_DATASETS` + `/api/maps/custom/[datasetId]`, Custom empty upstream again, endpoints constrained to Nexus API paths so the proxy can't forward a Bearer token off-host. Zen keeps the layer via config, no ABI patch.
2. **Invite send-failure bypassing cleanup** — one guarded `issue_and_send_invite_sign_in` used by both routes and both agent tools; existing-member invite resends idempotently instead of 400.

I've left both threads open until #1131 lands.

### Two branch-level findings worth knowing

Neither is caused by #1131 — both reproduce on this tip:

- `pytest app/services/ app/core/` → 391 passed, **4 failed**: `chat/service_test.py::test_complete_chat_request_with_provider_uses_completion`, `coding_environment/…::test_duplicate_name_conflict`, and two in `providers/service_test.py`.
- `vitest run` → 232 passed, **3 failed**: `graph-query/explore-state.test.ts` (2) and `triples-export.test.ts` (1). Separately, `src/lib/feature-access.test.ts` is written against `node:test`, so vitest reports it as a failed suite with 0 tests even though all 10 assertions pass under node — worth either porting it or excluding it, otherwise `vitest run` can never be green.

Issue #1127, opened 2026-07-30
Improve ABI onboarding: local-first defaults via a new ollama marketplace module (Phi-3.5) + docs site with use-case runbooks


Dr0p42:
## Goal

Make a brand-new ABI project **work out of the box with zero API keys**, and give users a documentation site that shows *how to solve real problems* with ABI — not just an API reference.

Two workstreams, one outcome: someone clones/creates an ABI project and gets a working agent + a runbook to follow, without a credit card.

---

### Current state (the problem)

A fresh `abi new project` is **cloud-only and key-gated**. All three generated configs enable OpenAI + OpenRouter:

- `libs/naas-abi-cli/naas_abi_cli/cli/new/templates/project/config.yaml:56-64` → `naas_abi_marketplace.ai.chatgpt` (`OPENAI_API_KEY`) + `naas_abi_marketplace.ai.openrouter` (`OPENROUTER_API_KEY`)
- same in `config.local.yaml:106-111` and `config.remote.yaml:40-45`
- `config.yaml:77-79` → `default_chat_model: "claude-opus-4.8"`, `default_embedding_model: "text-embedding-3-large"`
- `config.yaml:17-25` → opencode providers `openrouter` + `openai`, both requiring keys
- `global_config.ai_mode: "cloud"` (`config.yaml:14`)

So the first-run experience is: scaffold a project → nothing works → go get two API keys.

There is also **no `naas_abi_marketplace.ai.ollama` module**. Ollama-backed models exist but are scattered across vendor-named modules, each hand-rolling `ChatOllama`:

- `ai/gemma/models/gemma3_4b.py` (`PROVIDER = "ollama"`)
- `ai/qwen/models/qwen3_8b.py`
- `ai/deepseek/models/deepseek_r1_8b.py`
- `ai/llama/models/llama_3_3_70b.py`

And there is **no Phi model anywhere** in the marketplace (`grep -rn "phi3\|phi-3\|phi4"` → 0 hits).

### What we want

1. **New marketplace module `naas_abi_marketplace.ai.ollama`** following the standard module layout (`__init__.py` with `ABIModule`/`Configuration`, `models/`, `agents/`, `ontologies/` — mirror `ai/gemma/__init__.py` and `ai/chatgpt/`).
   - Config surface: ollama base URL (default `http://localhost:11434`), optional list of models to pull/expose.
[...]

PR #1129, opened 2026-07-30
feat(onboarding): local-first defaults via new ollama marketplace module (Qwen2.5 3B + local embeddings)

Dr0p42:
Implements Workstream A of #1127: a new ABI project boots with **zero API keys**, on local models served by Ollama.

## The problem

A fresh `abi new project` was cloud-only and key-gated:

- generated configs enabled `ai.chatgpt` + `ai.openrouter`
- defaults were `claude-opus-4.8` (chat) and `text-embedding-3-large` (embeddings)
- `uv add` pulled the `ai-chatgpt` extra
- `global_config.ai_mode: "cloud"`

So the first-run experience was: scaffold → nothing works → go get two API keys.

### New `naas_abi_marketplace.ai.ollama` module

| Canonical id | Ollama tag | Role |
|---|---|---|
| `qwen-2.5-3b` | `qwen2.5:3b` | default chat model **and** tool-using agents (32k ctx, ~1.9GB) |
| `nomic-embed-text` | `nomic-embed-text` | default embedding model (768 dims, ~274MB) |

Plus provider factories for `ollama`, so any tag works off-catalog without a model file; an `OllamaAgent`; and a module README.

**Why a local embedding model matters:** every other marketplace `EmbeddingModel` is cloud-backed (OpenAI/Bedrock), so before this a keyless project had no working vector store.

**Why the default chat model must emit structured tool calls.** `AbiAgent` and `OntologyEngineerAgent` bind tools, and both default to `claude-sonnet-5` upstream — which no longer resolves with the cloud modules off, so **the API failed to boot** with `ModelNotFoundError`. They are pinned to the local default instead, which means that default has to work for tools too. Qwen2.5 3B does, so one ~1.9GB model covers chat and agents alike.

Models evaluated and rejected for this slot, recorded in both READMEs so nobody repeats the experiment:

| model | why not |
|---|---|
| `qwen2.5-coder:3b` / `:7b` | Advertise `['completion','tools','insert']` and pick the right function and arguments, but emit them as bare JSON in the message body instead of inside the `<tool_call>` tags their own template declares — so Ollama never fills `tool_calls` and LangGraph sees plain text. **0/3** at both sizes vs **3/3** for `qwen2.5:3b`, identical on langchain-ollama 0.3.10 and 1.1.0. Supporting them would mean hand-parsing tool calls out of message content; that is Ollama's job, and the shim would be permanent maintenance, so it was rejected. |
| `phi3.5` | `['completion']` only — no tool support. |
| `gemma3:4b` | `['completion','vision']` — no tool support. |

The rule the READMEs state: a candidate earns this slot by passing a real `bind_tools` check, because Ollama's `capabilities` list is not a reliable signal.

### Cross-platform endpoint resolution (`endpoint.py`)

Ollama is a local server, so "where is it?" is a platform question. Rather than assume an address, the module builds an ordered candidate list and uses the first that answers `GET /api/tags`: explicit config → `ABI_OLLAMA_BASE_URL` → `OLLAMA_HOST` → loopback → (WSL only) the Windows host from `/etc/resolv.conf`.

| Platform | Behaviour |
|---|---|
| **macOS** | `localhost`; binary also looked up in the `.app` bundle and Homebrew paths |
| **Linux** | `localhost`; never spawns a second server, since systemd may already own the port |
| **Windows WSL** | Both topologies — ollama inside the distro, *or* the Ollama Windows app on the host, which plain `localhost` never reaches under NAT networking |

This follows the NAT-vs-mirrored precedent already in `naas_abi_cli.cli.dev` (`BIND_HOST`/`BROWSER_HOST`). Unreachable ollama logs a platform-specific hint and lets the project boot — the failure surfaces at first model call, not at startup. Every platform signal is injectable, so all three paths are unit-tested from any machine.

### Project template

- `ai_mode: local`; ollama is the only AI provider module enabled
- `model_registry` defaults point at `phi-3.5` + `nomic-embed-text`
- `uv add` installs the `ai-ollama` extra
- cloud providers documented as opt-in, with the extra to add
- compose gets `host.docker.internal:host-gateway` + `ABI_OLLAMA_BASE_URL`, so containerised deployments reach the host's ollama on Linux Docker too
- a README for the generated project (the template was an empty file)

## Sharp edge worth knowing

`config.yaml` is rendered as a Jinja template **before** it is parsed as YAML, so a `{{ secret.X }}` inside a `#` comment is *still* resolved — which prompts for a key interactively, or hard-fails with no TTY. My first attempt at "commented-out cloud examples" reintroduced the very prompt this PR removes. The examples now use a `SECRET_REF` placeholder, and the generated README carries the real snippet with an explanation.

## Verification

End to end from a throwaway directory (`/tmp`), with `OPENAI_API_KEY`, `OPENROUTER_API_KEY` and `ANTHROPIC_API_KEY` explicitly unset:

- `abi new project` → `abi config validate` → **Configuration is valid**
- `abi dev up` → oxigraph **ready**, api **ready**; all agents exposed (`Abi`, `DemoLocal`, `Ollama`, `Ontology_Engineer`)
- engine resolves `qwen2.5:3b` (chat) + `nomic-embed-text` (768 dims), endpoint auto-detected
- real completion returned; embeddings returned a 768-dim vector
- the same model produced a real structured `tool_call` when tools were bound
- generated configs contain **no** cloud key references; generated `.env` has no keys

Checks:

- 38 new tests for the module (endpoint resolution across all three platforms, on_load registration, keyless config)
- `ruff` clean; `mypy -p naas_abi_marketplace.ai.ollama` and `mypy -p naas_abi_core` clean
- Regression baseline: ran the marketplace-AI + core suites against HEAD and against this branch. Failure/error lists are **byte-identical** (15 failed / 37 errors, all pre-existing and environmental — missing `.env`, uninitialised modules, agent tests making real calls); this branch adds passing tests only.

## Pre-existing issues found, not fixed here

- `test_setup_local_deploy_hardens_fuseki_for_reliability` fails on `main`: the assertion expects one `image:` line in the fuseki→yasgui region, but HEAD's own template has two (a `curlimages/curl` init service sits between them).
- `abi config validate` reports valid even when a configured module isn't installed and the `model_registry` defaults cannot resolve — the real check only happens at engine boot.

## Follow-ups still open on #1127

Workstream B (docs site + use-case runbooks) is untouched, including the broken docs deploy (`docs.yml` copies from the gitignored `docs/.quartz_conf/`).

Closes nothing yet — #1127 stays open for Workstream B.


Dr0p42:
Swapped the default chat model from Phi-3.5 to **Qwen2.5 3B** per review.

Phi-3.5 is completion-only in Ollama, which had forced a *second* model (`llama-3.2`) into the default setup purely to keep `AbiAgent`/`OntologyEngineerAgent` working. Qwen2.5 3B advertises the `tools` capability, so one small model now serves the default chat model **and** both tool-using agents.

Net effect on onboarding: **two model pulls instead of three**, and less to download — `qwen2.5:3b` is ~1.9GB vs `phi3.5`'s ~2.2GB.

Re-verified from a throwaway dir with all three cloud keys unset: `abi new project` → `abi dev up` reaches **api ready** (all six agents exposed), engine resolves `qwen2.5:3b` + `nomic-embed-text`, real completion returns, embeddings return a 768-dim vector, and the same model produces a real `tool_call` when tools are bound. 38 module tests pass; ruff and mypy clean.

Dr0p42:
Switched `default_chat_model` to **Qwen2.5-Coder 3B** as requested.

One thing that came out of testing and shaped the design: **the coder models don't produce structured tool calls**, at any size. `ollama show` advertises `['completion','tools','insert']` for them, and the model does pick the right function and arguments — it just emits them as bare JSON in the message body instead of inside the `<tool_call>` tags its own chat template declares, so Ollama never fills the `tool_calls` field and LangGraph sees a plain text reply.

Measured via `bind_tools` (the path agents use), identical prompts:

| model | size | structured tool calls |
|---|---|---|
| `qwen2.5-coder:7b` | 4.7GB | 0/3 |
| `qwen2.5-coder:3b` | 1.9GB | 0/3 |
| `qwen2.5:3b` | 1.9GB | 3/3 |

So Coder is the default chat model, and `abi_agent_model`/`ontology_engineer_model` stay on the general Qwen2.5 3B — otherwise `AbiAgent` and `OntologyEngineerAgent` are silently broken. Both configs carry an inline warning not to collapse them onto the coder model.

Cost: three pulls (~4.1GB) vs two (~2.2GB) for a single general model. Happy to go back to one model if the extra download outweighs a code-tuned default.

Re-verified from a throwaway dir, all cloud keys unset: `abi new project` → `abi dev up` reaches **api ready** (six agents), engine resolves `qwen2.5-coder:3b` + `nomic-embed-text`, real chat and code completions, 768-dim embeddings, and the pinned agent model produces a real structured `tool_call`. 39 module tests pass; ruff clean.

Dr0p42:
Reverted the Qwen2.5-Coder change (`00437e41c`). Decision: use a model that genuinely supports tool calling, with no custom handling.

Final design is back to **one local chat model** — general Qwen2.5 3B for both chat and the tool-using agents — plus nomic-embed-text. Two pulls (~2.2GB), stock `langchain-ollama`, zero custom code.

The measurements are kept in both READMEs as an "evaluated and rejected" table (qwen2.5-coder 3b/7b, phi3.5, gemma3:4b), with the rule that a candidate for the default-chat slot must pass a real `bind_tools` check — Ollama's `capabilities` list advertises `tools` for the coder models and is not trustworthy. That keeps the finding without carrying the model.

Code is byte-identical to the state verified end to end earlier in this PR (fresh project, no API keys, `abi dev up` → api ready, real completion, 768-dim embeddings, real structured tool call); only the two READMEs changed. 38 module tests pass, ruff and mypy clean.

Dr0p42:
Followed up on "will a 3B model actually work as an agent?" — the earlier 3/3 was one trivial tool, which only proves the plumbing.

Retested with **8 plausible tools**, scoring whether the *right* tool is picked and whether the model abstains when none applies:

| model | correct routing |
|---|---|
| `qwen2.5:3b` | **8/8** (including correctly not calling a tool for "write a haiku") |
| `qwen2.5:1.5b` | 6/8 — twice answered in prose instead of calling a tool |

So the default is genuinely competent at routing, not just the easy case.

This also **corrected a claim I'd made**: I had described `qwen2.5:1.5b` as an equivalent lighter option because it advertises tool support. It's measurably worse, so the READMEs and model docstring now present it as a constrained-hardware fallback with the numbers attached (`8d39addcf`, docs only).

Still unmeasured and not claimed anywhere: multi-step tool chains, argument correctness under long system prompts, and the full AbiAgent toolset over a real conversation. Those are where a 3B model is most likely to disappoint, and they'd need a real agent-level harness to judge.

Dr0p42:
Ran the multi-tool / agent-level test. Driven through the **real `Agent` loop** (LangGraph, tools actually executed, scored on what ran — not on what the model claimed).

**15/19 checks passed. All four failures are the same thing: multi-step chaining.**

| group | result |
|---|---|
| argument accuracy, incl. a typed `int` | 6/6 ✅ |
| routing under a 3,110-char system prompt | 4/4 ✅ |
| multi-turn with history | 2/2 ✅ |
| abstention (no phantom tool calls) | 2/2 ✅ |
| **two-step chain, 9 tools bound** | **0/4** ❌ |

Failure mode, deterministic across attempts: asked for "the budget of the team Alice belongs to", it calls `find_employee_id` with `{"sparql": "?x foaf:name \"Alice\"."}` — a parameter belonging to a *different* tool. Schema validation rejects it, and the agent then apologises in prose instead of recovering.

**Isolated the cause: tool count, not one confusable tool.** Same chain, same prompt:

| tools bound | chained correctly |
|---|---|
| 2 | 2/2 |
| 3 | 2/2 |
| 4 (incl. the tool whose parameter leaks) | 2/2 |
| 9 | 0/2 |

**Why this matters for what we ship** — measured on a fresh project:

| agent | tools | sub-agents | prompt |
|---|---|---|---|
| `AbiAgent` | **9** (5 utility + 4 `transfer_to_*`) | 4 | 2,263 chars |
| scaffolded project agent | 5 | 0 | 555 chars |
| `OllamaAgent` | 5 | 0 | 851 chars |

`AbiAgent` binds exactly 9 and delegating via `transfer_to_*` is inherently multi-step — so **Abi's supervisor behaviour is the weak spot on the local default**. The 5-tool agents are comfortable. Both READMEs now state this with the numbers and point at `abi_agent_model` as the escape hatch (`6575db509`, docs only).

Also correcting myself: I earlier said the scaffolded agent "binds no tools". The framework injects 5 utility tools (`get_time_date`, `write_file`, `read_file`, `list_dir`, `run_terminal`) even when the class declares `tools: list = []`.

**Open question for you:** ship as-is with this documented, or test a bigger local model (e.g. `qwen2.5:7b`, ~4.7GB) for `abi_agent_model` so Abi chains reliably? That's a ~2.5GB extra download on first run. Happy to measure it before you decide.

Dr0p42:
Wired the Nexus API to the module — Workstream A item 3 ("Wire the runtime") plus the cross-platform half of it.

## The gap

The local-first work stopped at the engine. The Nexus API carried its own older Ollama layer that never learned about the module, and the two disagreed where a user would notice:

- **Wrong model.** `services/ollama.py` hardcoded `qwen3-vl:2b` as the model to prefer, pull and recommend. Follow the generated README, pull `qwen2.5:3b`, open Nexus chat → *"Install Ollama and run: `ollama pull qwen3-vl:2b`"*. A model the project never installs and nothing else references.
- **Wrong host.** `localhost:11434` was hardcoded in **9 places**. Under `config.local.yaml` the module talks to `host.docker.internal`, but `/api/ollama/status` probed `localhost` *inside the container* and reported Ollama offline while the engine was using it fine. Same failure under WSL with the Ollama Windows app on the host — which the module already solves.
- `find_ollama()` branched on Darwin/Linux only, no WSL awareness.

## The fix — one source of truth

- New `defaults.py` in the ollama module holds the chat/embedding tags and the fallback preference order. No third-party imports, so the API can read it **without the `ai-ollama` extra**; `langchain_ollama` moved to a lazy import inside the provider factories to keep that true (subprocess-isolated test asserts it).
- The model files take `MODEL_ID` from those constants, so the catalog and the API cannot drift apart again.
- `services/ollama.py` gains `resolve_endpoint()`, delegating to the module's already-tested resolver (`ABI_OLLAMA_BASE_URL` → `OLLAMA_HOST` → WSL host → loopback). Cached, since resolution probes the network and this is per-request. `find_ollama()` delegates too, which is how it gains WSL awareness.
- Every `endpoint` argument now defaults to `None` and resolves per call instead of binding a literal at import — so the environment can actually override it.

The vision path keeps `qwen3-vl:2b`: an image request needs a vision model, which is a genuinely different question from "which model does a keyless project ship".

Verified the container case end to end — with `ABI_OLLAMA_BASE_URL=http://host.docker.internal:11434`, `get_ollama_status()` now returns that endpoint and `qwen2.5:3b`.

## Tests

10 new tests that assert the two layers **agree**, rather than asserting literals (which would just re-encode the drift): same tag on both sides, container override honoured, resolution cached, a broken resolver degrades to the default instead of taking the API down, no `endpoint` argument defaults to a literal, and a scan that fails on any newly introduced hardcoded `11434` URL anywhere in the API.

**389 passed** across the nexus services + marketplace ollama selection. 3 failures remain (2 provider-logo, 1 chat-completion) — they reproduce identically on `6575db509` in a detached worktree, so they are pre-existing and untouched by this change. Pre-commit hooks clean: ruff, mypy (*no issues in 691 source files*), bandit.

---

Workstream A is now complete and self-consistent. Still open, both yours to call:

1. `qwen2.5:7b` for `abi_agent_model` so Abi chains reliably (~2.5GB extra on first run) — happy to measure before you decide.
2. Workstream B (docs site + runbooks). Worth noting the deploy is confirmed broken independently of this PR: `.gitignore:133` ignores `docs/.quartz_conf/` and `git ls-files` returns **zero** tracked quartz files, so the four `cp docs/.quartz_conf/…` steps in `docs.yml` have nothing to copy on a clean checkout. I would do that as its own PR.

libs/naas-abi-cli/naas_abi_cli/cli/deploy/templates/local/docker-compose.yml
```diff
@@ -52,6 +52,11 @@ x-abi-template: &abi_template
     # Cached volumes for performance
     - venv_cache:/app/.venv       # Python virtual environment cache
     - huggingface_cache:/root/.cache/huggingface  # Docling / HF model weights
+  # Ollama serves the default chat + embedding models and runs on the *host*,
+  # not in the stack. `host.docker.internal` is built in on Docker Desktop
+  # (macOS/Windows) but not on Linux Docker, where host-gateway supplies it.
+  extra_hosts:
+    - "host.docker.internal:host-gateway"
```
Dr0p42, line 71: **[P1] Make host Ollama reachable from Linux containers**

`host-gateway` only provides routing/name resolution; it does not change what address Ollama listens on. The standard Linux installation binds to `127.0.0.1:11434`, so a container connecting to the host's bridge address through `host.docker.internal` will get connection refused. A fresh Linux project following the generated README can therefore boot but fail on its first chat or embedding call.

Please either run Ollama as a Compose service or configure and document a safe host bind address. Ollama documents the loopback-only default here: https://docs.ollama.com/faq#how-can-i-expose-ollama-on-my-network

Dr0p42 (commented):
Three issues affecting the advertised local-first behavior: Linux Docker reachability, the Ollama agent's effective tool surface, and the configured context size.

libs/naas-abi-marketplace/naas_abi_marketplace/ai/ollama/agents/OllamaAgent.py
```diff
@@ -0,0 +1,84 @@
...
+
+<operating_guidelines>
+- Maintain a clear, concise, and professional tone.
+- Format responses as clean, well-structured Markdown.
+- Prefer short, direct answers; offer to go deeper rather than pre-empting.
+</operating_guidelines>
+
+<constraints>
+- Preserve the language of the user's message in your response.
+- You have no tools and no network access. If a task needs live data or an
```
Dr0p42, line 37: **[P2] Align the prompt with the agent's actual tool surface**

`Agent` defaults `enable_default_tools=True`, and this constructor does not override it. At runtime this agent receives `get_time_date`, `write_file`, `read_file`, `list_dir`, and `run_terminal`, despite the prompt saying it has no tools or network access. This can make the model refuse available functionality; in a coding workspace, `run_terminal` also makes the privacy/external-action assertion inaccurate.

Please either pass `enable_default_tools=False` when constructing this agent, or update the prompt and privacy claims to accurately describe the injected tools.

libs/naas-abi-marketplace/naas_abi_marketplace/ai/ollama/models/qwen2_5_3b.py
```diff
@@ -0,0 +1,54 @@
...
+        name="Qwen2.5 3B",
+        owner="alibaba",
+        description=(
+            "Alibaba's Qwen2.5 3B running locally via Ollama. Small enough for "
+            "consumer hardware, 32k context, and supports tool calling — so it "
+            "can back both chat and tool-using agents with no API key."
+        ),
+        image="https://naasai-public.s3.eu-west-3.amazonaws.com/logos/ollama_100x100.png",
+        context_window=32768,
+        model=ChatOllama(
```
Dr0p42, line 48: **[P2] Configure the advertised context size**

`context_window=32768` is only ABI metadata; this `ChatOllama` instance leaves `num_ctx=None`, so Ollama uses its default 4096-token context. Long conversations and tool-heavy prompts will therefore truncate well before the advertised 32k window.

Please set `num_ctx=32768` (or configure `OLLAMA_CONTEXT_LENGTH`) or describe 32k only as a supported maximum requiring explicit configuration. Ollama documents the 4096 default here: https://docs.ollama.com/faq#how-can-i-specify-the-context-window-size

libs/naas-abi-cli/naas_abi_cli/cli/deploy/templates/local/docker-compose.yml
Dr0p42: Confirmed and fixed in 31636fe38. You're right that `host-gateway` only supplies routing and name resolution — the stock Linux install binds `127.0.0.1:11434`, so the container resolves the bridge address and then gets connection refused.

I went with the documented bind rather than a Compose service. Running Ollama in Compose would mean no GPU on macOS (Docker Desktop can't pass through Metal), a second copy of the model weights inside a volume, and it would break the "Ollama you already have on your machine" story that the rest of this PR is built on. Happy to revisit if you'd rather have the self-contained stack.

So:

- **`endpoint.py` gained `in_container()`** (checks `/.dockerenv`, `/run/.containerenv`, `/proc/1/cgroup`, all injectable) and `install_hint()` gained a Linux-container branch. It's checked *before* the WSL branch, since Docker Desktop's WSL backend is both and the bind address is the actionable problem there too. The hint now names the actual fix instead of "install from ollama.com" — which was actively misleading, since Ollama *is* installed and running.
- **Documented in three places**: the compose template (right at `extra_hosts`, where someone debugging will be looking), the module README, and the generated project README.
- **The exposure is called out, not glossed.** `OLLAMA_HOST=0.0.0.0` opens the port on every interface, so the docs say "trusted network only" and give the narrower `172.17.0.1:11434` bridge-bind alternative plus a firewall note. There's a `docker run … curl` one-liner to verify.

Worth noting for scope: `abi dev up` runs natively and is unaffected — this is the `abi deploy local` path. The generated README now says which is which.

7 new tests cover container detection and the hint (content, exposure warning, precedence over both the plain-Linux and WSL branches, and that macOS containers keep the macOS hint).

libs/naas-abi-marketplace/naas_abi_marketplace/ai/ollama/agents/OllamaAgent.py
Dr0p42: Correct, and this is the sharper version of something I'd already got wrong earlier in this PR — I claimed this agent "binds no tools" on the strength of `tools: list = []`, then had to correct myself when I measured the real agent tool counts. You've found where that same mistake was still live in the code rather than just in my description of it.

Fixed in 31636fe38 by passing **`enable_default_tools=False`**, not by rewriting the prompt. Two reasons: this is a general local assistant that has no use for `write_file`/`run_terminal`, and the privacy claim is the entire point of the agent — "nothing leaves this machine" reading as marketing copy while `run_terminal` sits in the tool list is worse than either alternative. Your point about the coding workspace is exactly right.

It also helps on a second axis: I measured earlier in this PR that `qwen2.5:3b` chains reliably at 2–4 bound tools and fails at 9. Dropping this agent from 5 to 0 keeps it well inside the range that works.

Tested by **constructing the agent and asserting `_tools == []`**, plus a specific check that `run_terminal` isn't present — I'd written a source-inspection test first (`assert "enable_default_tools=False" in getsource(...)`), which passes without proving anything about runtime, so I replaced it.

libs/naas-abi-marketplace/naas_abi_marketplace/ai/ollama/models/qwen2_5_3b.py
Dr0p42: Confirmed against a live server, and fixed in 31636fe38.

Before:
```
NAME          SIZE      PROCESSOR    CONTEXT
qwen2.5:3b    2.2 GB    100% GPU     4096
```

After passing `num_ctx`:
```
NAME          SIZE      PROCESSOR    CONTEXT
qwen2.5:3b    3.2 GB    100% GPU     32768
```

I set `num_ctx` rather than downgrading the claim, because `context_window` isn't only documentation — it's what ABI reasons with, so a 32k metadata value against a 4k allocation means the framework believes it has headroom it doesn't have, and Ollama truncates without an error. `CONTEXT_WINDOW` is now declared once on the class and passed to both the metadata and the client, so they can't drift apart again.

The cost is real and now documented in both READMEs: **~1GB extra resident** (2.2 → 3.2GB), with `num_ctx` / `OLLAMA_CONTEXT_LENGTH` as the lever on a constrained machine.

---

**This turned up a worse one on the embedding side, which I'd documented wrong.** Chasing your point, I checked `nomic-embed-text` and it reports `nomic-bert.context_length = 2048` — both READMEs claimed 8k (Nomic's card advertises 8192 via RoPE scaling; the Ollama build doesn't expose it). And unlike the chat model, **`num_ctx` does not lift it**.

Measured — two ~3000-word documents differing *only in their final sentence*:

| | cosine |
|---|---|
| long docs, differing tails, default | **1.0** |
| long docs, differing tails, `num_ctx=8192` | **1.0** |
| same two tails, short enough to fit (control) | 0.868 |

Bit-identical embeddings, so the tail is dropped entirely — a long document gets indexed by its opening alone and nothing anywhere says so. That's a silent-corruption failure mode for the vector-store path, which is the one part of the keyless story with no cloud fallback. Corrected the claim everywhere and documented that chunking is mandatory. A test pins the description so the 8k figure can't come back.

---

While adding a test for the container override I also found that **an explicit endpoint was only a hint**: `resolve_base_url` probed candidates in order and took the first that answered, so pointing ABI at a remote GPU box and having that box go down silently rerouted to whatever local server was up — different machine, different models, no error. An explicit endpoint (module `base_url` or `ABI_OLLAMA_BASE_URL`) is now honoured as given, with the probe only reporting reachability. Auto-detection keeps first-that-answers, since finding one is its whole job, and `OLLAMA_HOST` stays a candidate rather than a pin because it's commonly set to bind a *server* rather than to target one.

Module tests: 38 → 59.

jravenel:
Rebased onto current `main` and resolved the `libs/naas-abi-cli/uv.lock` conflict (regenerated via `uv lock`). PR is mergable again.

Need to address (or confirm already handled) the `sqlite3.OperationalError: database is locked` bus contention on `python_queue` during concurrent first-boot schema load. Suggested direction: WAL + `busy_timeout` / retry. Tracking in a follow-up issue if you prefer not to block this PR.

After that, this is ready to squash-merge so #1136 can reshape onto local-first Ollama defaults.

jravenel:
When #1136 lands after this, root `config.yaml` becomes local-first. Docker with `ENV=local` / `ENV=remote` stays on its overlay files and should not break.

jravenel (commented):
## Second-pass review (post-rebase)

Checks are green and prior review items (Linux bind docs, `enable_default_tools=False`, `num_ctx=32768` on the registered Qwen model) look addressed. One new defect stands out from the Nexus endpoint wiring.

### [P1] WSL-resolved Ollama IPs are rejected by provider endpoint validation

`resolve_endpoint()` can return a WSL host address like `http://172.22.80.1:11434` (from `/etc/resolv.conf`). Chat then sets that on the provider config, and `validated_provider_endpoint()` calls `_validate_endpoint_url(..., allow_localhost=True)`.

Even with `allow_localhost=True`, private **IP literals** are still rejected unless the host is in `_LOCAL_HOSTS = {localhost, 127.0.0.1, ::1}`:

```python
if _is_ip_literal(host) and _is_private_or_local_ip(host):
    if not allow_localhost or host not in _LOCAL_HOSTS:
        raise UnsafeProviderEndpointError(...)
```

So the WSL topology this PR explicitly supports (Ollama Windows app + ABI in the distro) can auto-detect a reachable endpoint, then fail at chat time with an unsafe-endpoint error.

- `host.docker.internal` is fine (hostname, not an IP literal)
- `localhost` / `127.0.0.1` are fine
- WSL NAT gateway IPs are not

**Fix direction:** for `config.type == "ollama"`, allow private/link-local IPs (or trust `resolve_endpoint()` output and skip the private-IP deny list). Also align `stream_with_ollama_tools` (which currently bypasses validation) with the same rule.

Refs: `provider_runtime.py` (`_LOCAL_HOSTS`, `validated_provider_endpoint`), `chat/service.py` (`endpoint=resolve_endpoint()`), `endpoint.py` (`wsl_host_addresses`).

### [P2] Off-catalog Ollama chat factory omits `num_ctx`

`qwen2_5_3b.py` correctly sets `num_ctx=32768`, but the provider factory in `ai/ollama/__init__.py` builds bare `ChatOllama(model=..., base_url=..., temperature=0)`. `get_chat_model("qwen2.5:1.5b", provider="ollama")` (the documented constrained-hardware fallback) silently gets Ollama's 4096 default again. Pass `num_ctx` (or a shared default) through the factory.

### [P3 / product] `config.remote.yaml` template defaults to local Ollama

Intentional for keyless remote boots, but easy to ship a hosted deploy that expects cloud and gets a missing Ollama instead. README already hints at cloud for multi-user hosts; consider making remote default cloud-or-explicit, or a louder first-run warning.

### Not blockers

- SECRET_REF pattern in templates avoids the Jinja-comment secret footgun
- Compose `host-gateway` + Linux bind docs look correct after the earlier fix
- `OllamaAgent` privacy claim matches `enable_default_tools=False`
- Autostart / ensure-ready UI path remains commented; binary-first `ensure_ollama_ready` is less urgent while that stays dead

**Verdict:** fix the P1 before squash-merge; P2 can land in-PR or immediately after. Happy to re-check once the validation allowlist is widened for Ollama.

jravenel:
Fixed the P1 from the second-pass review.

- Ollama validation now allows private/link-local IPs (WSL gateway, LAN) via `allow_private_lan=True`
- Cloud metadata `169.254.169.254` still rejected
- Custom/cloud providers unchanged (still deny private LAN)
- `stream_with_ollama_tools` now uses `validated_provider_endpoint` too
- Tests: WSL IP, `host.docker.internal`, metadata reject, custom still denies private IP — **10 passed**

Ready for another look / squash-merge after CI.

PR #1131, opened 2026-07-31
fix(nexus): generic Custom Maps extension point + guarded invite sends (#1126 follow-ups)

Dr0p42:
Follow-up for the two review threads still open on #1126. Targets `zen-next-abi-20260730` so it lands inside that review vehicle rather than on `main`.

## 1. Zen-only Maps layer in upstream ABI — [P1 thread](https://github.com/jupyter-naas/abi/pull/1126#discussion_r3691303721)

`bbc334795` registered `ontologist-north-america` as a first-class Custom dataset in upstream ABI, against the Nexus `AGENTS.md` contract ("Custom … *(empty upstream)* … do not ship Zen-only datasets here"). Upstream also has **no** `/api/intelligence/ontologist-north-america` route, so the catalog entry always ended at a 502 in a normal ABI deployment, and its UI copy hard-coded Zen, Sanax and NaasAI.

The comment offered two ways out: move it to the Zen overlay, or add a generic configured extension contract. This takes the second — it removes the Zen specifics from upstream **and** leaves Zen a supported way to keep the layer without patching ABI.

- New `src/lib/maps-custom-datasets.ts`: `NEXT_PUBLIC_MAPS_CUSTOM_DATASETS` holds a JSON array of layer descriptors. Custom stays empty upstream, and a layer only appears where an operator pointed it at a backend that exists.
- New `/api/maps/custom/[datasetId]` replaces the Zen-specific proxy. The target comes from config keyed by dataset id, never from the request.
- `endpoint` must be a **Nexus API path**. Absolute and protocol-relative URLs are rejected at parse time, so the caller's Bearer token is never forwarded off-host. Ids must be route-safe, and a configured id cannot shadow a built-in dataset.
- New `MapsCustomFeed` renders any descriptor; middleware's anonymous-curl gate is now the `/api/maps/custom/` prefix rather than one hard-coded path.
- Deleted: `maps-ontologist-north-america.tsx`, `api/maps/ontologist-north-america/route.ts`, `ontologist-north-america.pins.json`.
- `AGENTS.md` documents the contract; `datasets.test.ts` now asserts Custom is empty upstream.

**Zen keeps the layer** by setting:

```json
NEXT_PUBLIC_MAPS_CUSTOM_DATASETS='[{"id":"ontologist-north-america","title":"Ontologist, North America","description":"…","icon":"Brain","order":0,"endpoint":"/api/intelligence/ontologist-north-america","fitMaxZoom":4}]'
```

## 2. Invite send-failure bypasses cleanup — [P2 thread](https://github.com/jupyter-naas/abi/pull/1126#discussion_r3691303729)

Both invite routes issued a challenge and sent it inline, skipping the invalidate-on-failure logic that already existed in `ensure_user_and_send_invite_email`. A raising `EmailService.send` returned 500 with an active challenge the invitee never received; the retry then hit `400 "User is already a member"`, so the admin had no way to resend.

- Extracted `issue_and_send_invite_sign_in` — invalidates the challenge whether delivery raises *or* reports `False` (no transport). `ensure_user_and_send_invite_email` now wraps it, so there is one guarded path.
- Org + workspace invite routes call it. So do the two `nexus_admin_tools.py` invite tools, which had the same inline pattern.
- Re-inviting an existing member now resends idempotently (`status: "already_member"` on the workspace route) instead of 400. It still 400s if no membership row can be resolved.
- `stores/organization.ts` replaces rather than appends the returned membership, so a resend cannot duplicate the row in the UI cache.

## Also verified (no code needed)

Five older threads on #1126 were left unresolved but are genuinely fixed at head — `workspace_id`-belongs-to-org in both the FastAPI adapter and `nexus_admin_tools.py`, CLI `--dry-run` deferring `build_client`, OTP `Field(default=6, ge=4, le=10)` / `ge=1`, and durable per-org role features via `organization_role_features` + migration `0040`.

## Checks

- `ruff check` on all changed Python: clean.
- `pytest` over `app/services/` + `app/core/`: **391 passed**, 4 failed — all pre-existing on the branch, in `chat`, `coding_environment`, `providers` (untouched here). The 9 new tests pass.
- `tsc --noEmit`: clean.
- `vitest run`: 232 passed, 3 failed — all pre-existing and untouched: `feature-access.test.ts` is a `node:test` file vitest cannot parse (10/10 pass under node), plus assertion failures in `graph-query/explore-state` and `triples-export`.
- The repo pre-commit hook (`make check`) fails on 2 pre-existing mypy errors in `naas_abi_core/.../ForgejoAdapter_test.py`, a package this branch does not touch, so the hook was bypassed for the commit.


jravenel:
**Incident note (do not merge this PR to `main` from this path)**

An accidental GitHub merges API call (intended as a dry-run) landed this PR's head onto `jupyter-naas/abi` `main` as merge commit [`23d8eb9`](https://github.com/jupyter-naas/abi/commit/23d8eb9fc4e155e5bdd8b7feb3fe5ed71468c954) ("dry-run").

- Revert PR: https://github.com/jupyter-naas/abi/pull/1135 (`git revert -m 1 23d8eb9`)
- This PR stays **draft** against `zen-next-abi-20260730`. Do **not** re-merge it onto `main` as part of cleanup.
- After #1135 lands, continue normal review into the Zen review vehicle, then the usual path onto `main`.

When this work eventually lands properly, Zen still needs `NEXT_PUBLIC_MAPS_CUSTOM_DATASETS` for the Ontologist North America layer (upstream Custom stays empty).

jravenel:
## Retarget complete

**What changed**
- Re-applied #1131 onto latest `main` (tip `74841b8`, post #1135 revert). A plain rebase was a no-op because the dry-run merge left the original commit as an ancestor of `main`; cherry-pick recreated the patch as `0dc4d9153`.
- Added `Dockerfile.release` build `ARG` for `NEXT_PUBLIC_MAPS_CUSTOM_DATASETS` so deployments can bake Custom Maps descriptors into the Next bundle.
- Base retargeted: `zen-next-abi-20260730` → **`main`**.
- PR Checks (3.11 / 3.12 / container-scan) are green. Marked ready for review.

**Zen env (required before/with merge)**
Upstream no longer ships Ontologist, North America. Zen tip commit `5338b8b` wires:
- `.env.gcp.example` + `terraform/environments/zen/gcp.env.patch`: `NEXT_PUBLIC_MAPS_CUSTOM_DATASETS` JSON for the Ontologist layer (`endpoint`: `/api/intelligence/ontologist-north-america`)
- `docker-compose.gcp.yml`: passes the value as a **build arg** (and runtime env)
- Docs / Ontologist tool pointer updated to `/api/maps/custom/ontologist-north-america`

On the GCP VM: copy the new env line into `.env`, then rebuild `nexus-web` when deploying a build that includes this PR. Without that env, the Custom bucket stays empty.

**Not done**
- No merge (waiting on Jeremy).
- Zen ABI submodule pin left on prior tip; bump after merge.

PR #1148, opened 2026-08-03
feat(agent): progressive tool disclosure via on-demand search_tools

jravenel:
## Abi context-engineering optimization

Cuts Abi's per-turn prompt by not binding its full leaf-tool surface up front. When enabled, only an always-on core is bound to the model each turn — sub-agent delegation, workspace tools, and a new `search_tools` tool — and the rest are **deferred** and loaded on demand.

### How it works
- `search_tools(query)` does keyword matching over the deferred tools and reports the matches back (also supports `select:<name>`).
- The runtime derives the "loaded" set from the **transcript** (names surfaced by prior `search_tools` results), so the active tool set is a pure function of message history — stable across turns and reconstructions.
- Deferred-but-unloaded tools stay out of the prompt but remain **executable** via `_tools_by_name` once loaded.

### Placement (API-first)
Core agent runtime (`naas_abi_core/services/agent/`), so it flows through the `/agents/*` API; CLI and Nexus inherit it with no client changes.

### Safety
- **Off by default.** `AgentConfiguration.progressive_tool_disclosure=True` or `ABI_PROGRESSIVE_TOOLS=1`.
- Only activates with ≥5 deferrable tools (small sub-agents unaffected).

### Measurements (`qwen-2.5-3b`, CPU, "What can you do?")
| | 1st-call prompt | Wall clock |
|---|---|---|
| Baseline (full tools) | 4,225 tok | 73s |
| Progressive on | **2,545 tok (−40%)** | 54s |

**Mechanism validated end-to-end** (stream on an explicit task):
```
tool_usage: search_tools
tool_response: Loaded 5 tool(s) ... [loaded_tools] list_workspace_members, list_workspaces, create_workspace, ...
call_model: Abi        # discovered tools now bound and callable
ai_message: ...
```

### Honest caveats / follow-ups (not in this PR)
1. The −40% is **capped** because `AbiAgent`'s system prompt still text-lists every tool in `[TOOLS]` (separate from bound schemas). Gating that listing to the loaded/core set — and instructing the model to `search_tools` for the rest — should push the prompt toward the ~1,384-token floor and remove the "advertised but unbound" inconsistency.
2. On the 3B CPU model, the orchestrator does **not** reliably choose to call `search_tools` autonomously (it can emit an empty turn), and the meta-question "What can you do?" needs a deterministic capability answer rather than tool reduction. Works reliably when instructed and/or on a stronger model.

Retains the `num_predict` generation cap as an independent safety net.

<div><a href="https://cursor.com/agents/bc-b3436ea0-90ab-4340-9379-d4b6c03f199c?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-b3436ea0-90ab-4340-9379-d4b6c03f199c&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/libs/naas-abi-core/naas_abi_core/services/agent/Agent.py
+++ b/libs/naas-abi-core/naas_abi_core/services/agent/Agent.py
@@ -75,6 +75,10 @@ from naas_abi_core.services.cache.CachePort import DataType
 from sse_starlette.sse import EventSourceResponse
 
 from .tools.default_tools import default_tools
+from .tools.tool_search import (
+    build_tool_search_tool,
+    extract_loaded_tool_names,
+)
 from .tools.utils import can_bind_tools
 from .tools.workspace_tools import REQUIRES_WORKSPACE_KEY
 
@@ -347,6 +351,12 @@ class AgentConfiguration:
     system_prompt: str | Callable[[list[AnyMessage]], str] = field(
         default="You are a helpful assistant. If a tool you used did not return the result you wanted, look for another tool that might be able to help you. If you don't find a suitable tool. Just output 'I DONT KNOW'"
     )
+    # Progressive tool disclosure (Abi context-engineering doctrine). When True
+    # and the agent has a large leaf-tool surface, only an always-on core
+    # (delegation + workspace + the search tool) is bound to the model; the rest
+    # are deferred and loaded on demand via the ``search_tools`` tool. Can also
+    # be enabled globally with the ABI_PROGRESSIVE_TOOLS=1 environment variable.
+    progressive_tool_disclosure: bool = False
 
     def get_system_prompt(self, messages: list[AnyMessage]) -> str:
         return (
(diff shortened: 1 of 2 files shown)


jravenel:
wont do now.

PR #1159, opened 2026-08-03
feat(dev): auto-login to Nexus and generate the admin password

Dr0p42:
## Why

`abi dev up` allocates ports per worktree, so **every worktree is a distinct browser origin** with its own `localStorage` — and therefore its own manual login. Running several worktrees in parallel means retyping credentials constantly. This signs the browser in automatically.

It also stops shipping a guessable admin password. The two go together: a generated password is only painless if you never have to type it.

## How auto-login works (and what it deliberately isn't)

The login page submits the **ordinary password flow** with the seeded credentials. It is *not* a dev-only endpoint that mints a token without credentials — that version becomes a full account-takeover the day the flag ships on, for essentially the same DX.

The safety property: **the credentials still have to be valid**. Worst case if this ever escaped to a real deployment is a login page trying a password no account has.

Four gates, any one of which is sufficient:

| Gate | Where |
|---|---|
| Credentials passed in the api child process env only — never written to `.env` | `dev.py` `_apply_autologin_env` |
| `NexusConfig` has no matching field and is `extra="forbid"` → unsettable from `config.yaml` | `naas_abi/__init__.py` |
| Blanked outside local/development, or when `auth_password_enabled` is off | `Settings.model_post_init` |
| Emitted only when both halves survived | `/api/auth/config` |

## UX details that make or break it

- **Signing out sticks.** A suppression flag in `localStorage` stops auto-login from looping you back in — otherwise "Sign out" is a no-op. Cleared only by a *successful* manual sign-in, so a typo can't undo a deliberate sign-out.
- **Escape hatches:** `?nologin=1` on the login page, and `abi dev up --no-autologin`. Keeps this page (and `e2e/auth.spec.ts`) workable.
- The spinner says *"Dev auto-login as admin@example.com…"* so it's never mysterious why you're signed in.

## Generated password

`secrets.token_urlsafe(24)`, matching the Nexus seed's own `_generate_password()` rather than inventing a second style. Generated once per project and persisted to `.env`. **Existing values are always preserved**, so existing stacks keep working unchanged.

## Latent bug fixed along the way

The Nexus seed writes credentials back through python-dotenv `set_key`, which **quotes** them (`PASSWORD='admin'`). `dev.py` read `.env` values literally, so auto-login would have submitted the quotes too — looks right in the terminal, fails at the form. The same bug was latent in `_ensure_default_api_key_env`. Both now go through `_read_env_file()`, with regression tests.

Found by copying a *real* `.env` before booting; a `tmp_path` fixture would never have had the quotes.

## Verification

Ran against a **live stack** (`abi dev up`): `/api/auth/config` serves the credentials, and those credentials authenticate as the superadmin (`access_token` returned, `is_superadmin: true`).

- 35 CLI + 6 config + 9 auth-adapter pytest; 18 vitest
- `tsc --noEmit`, eslint (no findings in touched files), `uvx ruff`, mypy, `pnpm build`
- Full `make check` pre-commit hook passed

Not verified in a real browser — the Chrome extension wasn't connected here. The React wiring is covered by unit tests on the extracted decision logic plus a production build, but a human should eyeball the actual redirect once.

Pre-existing unrelated failure `local_test.py::test_setup_local_deploy_hardens_fuseki_for_reliability` — confirmed identical on a detached worktree of pristine `main`.


Dr0p42:
## Follow-up: the deploy path had the real hardcoded password

Checked whether `abi new project` also needed changing. It **didn't** — the scaffolded `.env` (`templates/project/.env`) carries no password at all, so a new project goes through `abi dev up`'s generator, already fixed in the first commit. Its `config.yaml` comment even claimed the password was *"auto-generated on first boot and written to `.env`"* — which was untrue until this branch, because `dev.py` wrote the literal `admin` before the seed could generate one. That comment is now accurate.

The actual hardcoded credential was one layer over, in **`abi deploy local`**:

```python
"NEXUS_USER_ADMIN_PASSWORD": "Admin1234!",   # deploy/local.py
```

rendered into the generated `.deploy/.env` — and published in the docs. So every local deployment anyone has ever made shares one well-known admin password.

It was also the odd one out: `POSTGRES_PASSWORD`, `MINIO_ROOT_PASSWORD`, `RABBITMQ_PASSWORD`, `FUSEKI_ADMIN_PASSWORD`, `ABI_API_KEY` and the Coder/Forgejo admin passwords in that same module are all already random. Fixed by moving it into the existing `RANDOM_ENV_KEYS` mechanism rather than inventing anything — generated once, preserved across `--regenerate` via `_ensure_env_var`, and existing deployments keep whatever their `.env` already has.

**Surfacing it matters more than generating it.** A generated password nobody is told is just a lockout, so it now prints in two places:
- `abi deploy local` prints the login block at the end — reading the value *back out of `.env`*, so a re-run over an existing deployment shows the real password rather than a freshly generated throwaway
- `abi stack start` prints it next to the login URL it already opens

Auto-login deliberately does **not** apply to this path — it's env-gated to `abi dev up` only, and a Docker stack can be reachable from other machines.

4 new tests: generated (and not `Admin1234!`), differs between deployments, survives `--regenerate`, and is actually printed.

Docs updated for the paths that describe `abi deploy local` (AGENTS.md, quickstart, nexus capability page, `config.local.yaml`). The nexus subrepo docs (README/START/CONTRIBUTING/TROUBLESHOOTING) are left alone — they document the standalone upstream `jravenel/nexus` repo, a different stack whose `Admin1234!` doesn't come from here.

Same verification as before: full CLI suite (216 passed), ruff, mypy, and the `make check` pre-commit hook.

PR #1167, opened 2026-08-03
fix(agent): stop the markdown formatting pass from erasing the answer

Dr0p42:
## Symptom

On a fresh `abi dev up` project: you talk to Abi, the abi process logs show the
LLM answered, but the Nexus UI stays on **"Processing…"** forever — no error, no
warning. Reported by @Dr0p42.

## Root cause

`Agent._pretty_display_markdown` runs a *second* LLM call to reformat the reply,
then assigned the result onto the response unconditionally:

```python
response.content = formatted_response.content.strip()
```

When that second call comes back blank, the real answer is silently replaced
with `""`. Captured live on `abi dev up` (Ollama `qwen2.5:3b`):

```
18:03:56.404  Model response: content='I am Abi, the orchestrator Agent developed by NaasAI...'
18:03:56.431  Applying Markdown pretty display to response
18:04:08.475  Markdown pretty display response:          <- empty
18:04:39.977  stream_invoke:2259 - Response:             <- empty
```

With an empty final state, `stream_invoke` emits no `ai_message` events, the SSE
stream carries no `content` frames, the assistant row is persisted empty, and
the client never leaves its `▌` placeholder. Nothing raises, so no error is ever
shown — which is exactly why this looks like a hang rather than a failure.

Why it comes back blank: the pass used `_chat_model_with_tools`. A tool-bound
model can answer a formatting prompt with a tool call and `content=''` — small
local models do this readily.

## Fix

- Keep the original response when the formatting pass returns blank.
- Route the pass through the unbound `_chat_model`. A pure text-reformat step
  has no business offering tools.

## Client hardening (same visible symptom, separate paths)

Found while reproducing, each independently reproducible:

- **Positional writes.** Streaming updated the assistant row via
  `updateLastMessage` (last element) rather than by id. The placeholder is not
  reliably last for the whole turn — `loadConversationMessages` appends local
  messages the server doesn't know by id *after* the API's — so the write could
  land on the wrong row and strand the placeholder. Now uses `updateMessageById`.
- **Stale `isStreaming`.** The outer error handler branched on `isStreaming`, a
  closure copy of React state that is still `false` there even though
  `setIsStreaming(true)` ran earlier in the same call. It appended a second
  bubble instead of replacing the placeholder — leaving a spinner that only a
  reload cleared. (The file already documents this trap for `streamingMessageId`
  a few lines up; the same fix was never applied here.) Now branches on whether
  a placeholder actually exists.
- **422 race.** Sending before `/api/agents/` resolves posts `agent: ""`, which
  the API rejects with `422 string_too_short`. On a freshly booted project this
  window is seconds wide (boot-time `/api/ontology/*` calls delay the agents
  fetch). Now waits for the list and surfaces a notice instead of firing a
  request that cannot succeed.

## Verification

- Two regression tests in `Agent_test.py`. Both **fail on `main`** with exactly
  the data loss (`assert '' == 'I am Abi...'`) and pass here.
- `Agent_test.py`: 19 passed. `test_agent_stream_invoke` fails in a full-file run
  — confirmed **pre-existing**, it fails identically on `main` and passes in
  isolation on both (live-LLM flake).
- `uvx ruff check` clean; `mypy` clean on the changed file.
- Frontend `tsc --noEmit` clean; the client changes were exercised by driving the
  real UI headlessly against a live stack (reproduced the stuck placeholder
  before, and the 422 path).

## Note for reviewers

The client changes make the placeholder impossible to strand, but the answer-
erasing bug is the one that produced the reported symptom. I'd take the
`Agent.py` change even if the frontend ones are debated.

Worth a follow-up (not in scope here): this formatting pass doubles latency on
every reply (a full extra round-trip — 12s in the trace above), and the local
SQLite hit `database is locked` under concurrent stream writes during testing.

### DIFF
--- a/libs/naas-abi-core/naas_abi_core/services/agent/Agent.py
+++ b/libs/naas-abi-core/naas_abi_core/services/agent/Agent.py
@@ -1027,7 +1027,11 @@ Reformat the input into clean, readable Markdown. Preserve all meaning and detai
         ]
 
         try:
-            formatted_response = self._chat_model_with_tools.invoke(prompt)
+            # Use the *unbound* model: this pass only reformats text, and a
+            # tool-bound model may answer a formatting prompt with a tool call
+            # and `content=''` (small local models do this readily), which used
+            # to wipe the real answer below.
+            formatted_response = self._chat_model.invoke(prompt)
             logger.debug(
                 f"Markdown pretty display response: {formatted_response.content}"
             )
@@ -1035,7 +1039,18 @@ Reformat the input into clean, readable Markdown. Preserve all meaning and detai
             if not isinstance(formatted_response.content, str):
                 return response
 
-            response.content = formatted_response.content.strip()
+            formatted = formatted_response.content.strip()
+            if not formatted:
+                # Never let a failed formatting pass destroy the answer. This
+                # surfaced as an agent that "replied" in the logs while the UI
+                # streamed nothing and sat on its placeholder forever.
+                logger.warning(
+                    f"Markdown pretty display returned empty content for agent "
+                    f"'{self._name}'; keeping the unformatted response."
+                )
+                return response
+
+            response.content = formatted
             return response
         except Exception as e:  # noqa: BLE001
             logger.warning(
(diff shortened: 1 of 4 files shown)


Dr0p42:
Not sure this is still the case, closing for now

PR #1179, opened 2026-08-10
feat(nexus): pass access token to app-html iframe embeds

jravenel:
Append JWT to same-origin /app-html/ URLs so private HTML apps can authenticate.


FlorentRvl:
@jravenel fixed with: https://github.com/jupyter-naas/abi/pull/1181

PR #1200, opened 2026-08-18
fix(nexus): point module asset URLs at the live API port

### LINKED ISSUE #1199: fix(nexus): module asset URLs ignore the allocated API port
## Current situation

Nexus builds absolute URLs for module-served assets (agent logos, app avatars) from `global_config.public_api_host`. The engine default is `localhost:9879` with no scheme. `public_api_host()` then prefixes `https://`, so the JSON returns:

`https://localhost:9879/modules/<module>/assets/public/<file>`

That is only correct when the API is actually bound to 9879 over TLS.

`abi dev up` does not leave the API on 9879. It allocates a per-worktree port (`9879` plus an offset, then the next free port in range). The process already knows that port (`ABI_PORT`). It already rewrites `PUBLIC_WEB_HOST` to the Nexus web port. It does not rewrite `PUBLIC_API_HOST` or `NEXUS_API_URL`.

The frontend treats an `http(s)://` logo URL as final. It does not prefix `getApiUrl()`. The browser requests the default host and gets a failed load.

A second, related gap: workspace marks now hide a 404 behind a letter (`WorkspaceMark`, #1198). Agent avatars and the chat empty-state still render a raw ``. A bad URL shows a broken-image glyph.

A third gap: `GET /logos/<file>` only serves `apps/api/public/logos` from the engine package. A project that keeps tenant marks under `storage/logos` has no overlay, so those paths 404 even when the file exists in the project tree.

## Target situation

- Module asset URLs use the origin the API process is bound to when the configured host is still the engine default.
- Loopback hosts get `http://`, not `https://`.
- An explicit public hostname in config (Docker / remote) is left alone.
- `abi dev up` exports `PUBLIC_API_HOST` and `NEXUS_API_URL` for the allocated API port, the same way it already exports `PUBLIC_WEB_HOST`.
- `GET /logos/<file>` checks `storage/logos` in the project root, then the engine package defaults.
- Agent avatars use the same letter-or-icon first, image after `onLoad`, stay on the fallback on error.

## Suggested fix

1. `public_api_host()`: if `ABI_PORT` is set and the configured host is the engine default (loopback + 9879, or empty), return `http://<browser-or-localhost>:<ABI_PORT>`. If the configured host has no scheme, use `http` for loopback and `https` otherwise.
2. `_launch_api` in `abi dev up`: set `PUBLIC_API_HOST` and `NEXUS_API_URL` to `http://{BROWSER_HOST}:{api_port}`.
3. Serve `/logos/{path}` from `Path.cwd() / "storage/logos"` first, then `apps/api/public/logos`.
4. Shared `getLogoUrl()` on the web client: prefix relative paths with the runtime API; rewrite leftover loopback:9879 absolute URLs to that API.
5. Agent avatar / empty-state: same `onLoad` / `onError` pattern as `WorkspaceMark`.

## Test plan

- `abi dev up` on a worktree whose allocated API port is not 9879: `GET /api/agents` logo URLs use that port over `http`. The file under `/modules/...` returns 200.
- Config sets `public_api_host` to a real `https://` hostname: URLs keep that host. `ABI_PORT` does not override it.
- Config omits `public_api_host` and `ABI_PORT` is unset: default remains loopback 9879, scheme `http`.
- `storage/logos/mark.png` is served at `/logos/mark.png`. An engine default such as `/logos/abi-logo.png` still 200s.
- Agent with a valid logo: fallback first, image after load.
- Agent with a 404 logo: fallback stays, no broken-image glyph.

jravenel:
## Summary
- Module asset URLs were built from the engine default (`localhost:9879` plus an `https://` prefix) even when `abi dev up` bound the API to another port.
- `public_api_host()` now uses `ABI_PORT` when config is still that default, and `http` for loopback. An explicit public hostname is left alone.
- `abi dev up` exports `PUBLIC_API_HOST` and `NEXUS_API_URL` for the allocated API port, the same way it already exports `PUBLIC_WEB_HOST`.
- `GET /logos/<file>` checks `storage/logos` in the project root, then the engine package defaults.
- Shared `getLogoUrl()` rewrites leftover loopback:9879 URLs to the runtime API. Agent avatars and the chat empty-state keep the icon until the image loads.

Closes #1199

## Test plan
- [ ] `abi dev up` on a worktree whose API port is not 9879: `GET /api/agents` logo URLs use that port over `http`
- [ ] File under `/modules/...` returns 200 in the browser
- [ ] Config sets `public_api_host` to a real `https://` hostname: URLs keep that host; `ABI_PORT` does not override it
- [ ] `storage/logos/mark.png` is served at `/logos/mark.png`; an engine default such as `/logos/abi-logo.png` still 200s
- [ ] Agent with a valid logo: icon first, image after load
- [ ] Agent with a 404 logo: icon stays; no broken-image glyph

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,6 +109,9 @@ storage/vectorstore/*
 
 # `abi dev` runtime artefacts (per-worktree instance/pid/log files)
 .abi/
+# Same artefacts when this repo is the consumer's `.abi` checkout.
+# `{consumer}/.abi/dev` is `{this-repo}/dev`.
+/dev/
 # Dagster local home dir created by `abi dev up`
 .dagster/
 # Local Docker stack snapshots (abi stack snapshot create)
--- a/libs/naas-abi-cli/naas_abi_cli/cli/dev.py
+++ b/libs/naas-abi-cli/naas_abi_cli/cli/dev.py
@@ -361,6 +361,11 @@ def _launch_api(
     # Keep config-templated frontend URLs in sync when they reference
     # {{ secret.PUBLIC_WEB_HOST }}.
     env["PUBLIC_WEB_HOST"] = f"{BROWSER_HOST}:{nexus_port}"
+    # Module asset URLs are built from public_api_host / NEXUS_API_URL. The
+    # engine default is localhost:9879. This process is on a worktree port.
+    api_origin = f"http://{BROWSER_HOST}:{spec.port}"
+    env["PUBLIC_API_HOST"] = f"{BROWSER_HOST}:{spec.port}"
+    env["NEXUS_API_URL"] = api_origin
     cmd = ["uv", "run", "python", "-m", "naas_abi_core.apps.api.api"]
     return _spawn(spec, cmd, _project_root(), env)
 
--- a/libs/naas-abi-cli/naas_abi_cli/cli/dev.py
+++ b/libs/naas-abi-cli/naas_abi_cli/cli/dev.py
@@ -361,9 +361,12 @@ def _launch_api(
     # Keep config-templated frontend URLs in sync when they reference
     # {{ secret.PUBLIC_WEB_HOST }}.
     env["PUBLIC_WEB_HOST"] = f"{BROWSER_HOST}:{nexus_port}"
-    # Module asset URLs are built from public_api_host / NEXUS_API_URL. The
-    # engine default is localhost:9879. This process is on a worktree port.
+    # Module asset URLs are built from public_api_host. The dotenv adapter
+    # prefers .env over process env, so PUBLIC_API_HOST / NEXUS_API_URL here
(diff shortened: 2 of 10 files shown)


jravenel:
Follow-up: dropped the frozen `localhost:9879` host list.

`9879` is only the start of the `abi dev` API port window, not a sentinel for "config was not rewritten". The resolver is now:

- `ABI_DEV_PUBLIC_API_ORIGIN` (set by `abi dev up` on the process) wins, so a `.env` public hostname cannot shadow the allocated port
- otherwise a public hostname in config is left alone
- otherwise loopback (any port) plus `ABI_PORT` uses the live bind

The web helper rewrites a loopback absolute URL only when its port is not the runtime API port. No hardcoded `9879`.

PR #1241, opened 2026-09-03
feat(dataset): adopt DuckLake backend

### LINKED ISSUE #1240: Adopt DuckLake as the dataset service backend
## Context

The dataset service currently stores Hive-partitioned Parquet through an in-process DuckDB adapter. Its port supports `create`, `describe`, `list`, `write(append|replace)`, `query`, and `drop`, but it does not provide real snapshot history, upsert, or a native JSON column type.

Usage is currently limited, so breaking the dataset port now is preferable to retaining incompatible or misleading semantics.

There is also an independent correctness bug: a second append to a partitioned dataset fails because the partitioned DuckDB `COPY` does not use the `APPEND` option. The generic adapter contract does not catch this because its only second write uses `mode="replace"`.

## Decision

- Make DuckLake the primary dataset backend.
- Adopt catalog-level integer snapshot IDs rather than the current per-dataset UUIDs.
- Add primary-key metadata, native JSON columns, upsert, and snapshot listing/time travel to the port.
- Keep the plain DuckDB/Parquet adapter only if it can implement the revised semantics honestly; otherwise remove it.
- Use a SQLite DuckLake catalog in the default `config.yaml` for the zero-infrastructure `abi dev` runtime.
- Use a PostgreSQL DuckLake catalog in CLI-created `config.local.yaml` files.
- Give DuckLake a dedicated PostgreSQL database named `ducklake`; do not share the `abi`, `abi_dagster`, or `nexus` databases.

### Immediate correctness fix

- Add DuckDB's `APPEND` option to partitioned Parquet writes.
- Add a regression test that appends multiple times into both existing and new partitions.

### Dataset contract

- Add `json` to supported column types with deterministic serialization and validation.
- Add `DatasetSpec.primary_key` and validate that key columns exist.
- Add `mode="upsert"` with defined behavior for null keys and duplicate keys within an incoming batch.
- Replace string/per-dataset UUID snapshot semantics with catalog-level integer snapshots.
- Add snapshot discovery and historical querying.
- Ensure raw SQL queries that reference multiple datasets observe one coherent catalog snapshot.

### DuckLake adapter

- Add `DatasetSecondaryAdapterDuckLake` behind `IDatasetPort`.
- Cover it with `DatasetSecondaryAdapterContract` plus tests for JSON, upsert, partition-key moves, deletes, replace, time travel, and concurrent writers.
- Implement upsert with `MERGE INTO`; DuckLake does not currently provide primary-key/unique constraints for `ON CONFLICT`.
- Implement bounded retries for retriable DuckDB/DuckLake transaction conflicts. Retry the complete transaction against fresh catalog state and use backoff with jitter.
- Expose configuration for catalog DSN and data path.

### Configuration defaults

Default `config.yaml`:

```yaml
services:
  dataset:
    dataset_adapter:
      adapter: "ducklake"
      config:
        catalog: "sqlite:storage/datastore/datasets.sqlite"
        data_path: "storage/datastore/datasets/"
```

CLI-created `config.local.yaml`:

```yaml
services:
  dataset:
    dataset_adapter:
      adapter: "ducklake"
      config:
        catalog: "postgres:postgresql://{{ secret.POSTGRES_USER }}:{{ secret.POSTGRES_PASSWORD }}@postgres:5432/ducklake"
        data_path: "storage/datastore/datasets/"
```

Update both the repository configs and the `abi new project` templates.

### PostgreSQL initialization

- Add a DuckLake database init script alongside the existing Dagster and Nexus scripts.
- Update both:
  - `docker/postgres/initdb/`
  - `libs/naas-abi-cli/naas_abi_cli/cli/deploy/templates/local/docker/postgres/initdb/`
- The rendered script must create `ducklake`, connect to it, ensure the `public` schema exists, and grant the configured PostgreSQL user the required privileges.
- Add CLI deploy/scaffold tests verifying that the DuckLake init script and matching catalog DSN are generated.
- Provide an idempotent bootstrap or documented upgrade command for existing PostgreSQL volumes. Docker entrypoint init scripts only run when a volume is first initialized.

### Operations

- Treat the SQLite catalog as stateful data in snapshot/restore workflows because inlined rows can live in the catalog until flushed.
- Ensure catalog metadata and the data path are backed up and restored as one coherent state.
- Define maintenance for data flush, adjacent-file merge/compaction, and snapshot expiration.
- Document that moving an existing catalog from SQLite to PostgreSQL is a metadata migration, not only a DSN change.
- Record the architectural decision in `docs/adr/`.

## Acceptance criteria

- Repeated partitioned appends preserve all rows.
- JSON values round-trip and are queryable using DuckDB JSON operators.
- Upsert updates matched keys, inserts unmatched keys, and handles partition-key changes without duplicates.
- Historical queries return the expected state for a catalog snapshot.
- Concurrent PostgreSQL writers do not lose acknowledged writes.
- SQLite write conflicts are retried within a documented bounded policy.
- Default `config.yaml` uses SQLite DuckLake.
- CLI-created `config.local.yaml` uses the dedicated PostgreSQL `ducklake` database.
- A fresh local deployment creates that database and grants the configured user access.
- Existing-volume upgrade instructions or automation are included.
- Dataset service documentation, configuration tests, adapter contract tests, and an ADR are updated.

Dr0p42:
## Summary

- replace the plain DuckDB/Parquet dataset adapter with DuckLake
- add native JSON, primary-key-driven upsert, integer catalog snapshots, snapshot listing, and coherent time-travel queries
- retry complete write transactions on fresh state with bounded exponential backoff and jitter
- configure SQLite catalogs by default and a dedicated PostgreSQL catalog for local Docker deployments
- add idempotent PostgreSQL initialization, backup and migration guidance, and an architecture decision record
- propagate the revised dataset contract through the Nexus API and web types

## Validation

- repository pre-commit quality gate: Ruff, mypy, Bandit, marketplace checks, and Nexus production build
- core dataset tests: 17 passed, 1 opt-in integration test skipped
- PostgreSQL concurrent-writer integration test: passed against PostgreSQL 16
- Nexus dataset tests: 17 passed
- CLI configuration and deployment tests: 16 passed
- Nexus TypeScript typecheck: passed

Closes #1240


Dr0p42:
Reviewed this against the needs of a dataset-backed ingestion pipeline (content-addressed ledgers, idempotent re-runs, registries re-registered on every module load). Replacing the adapter rather than patching the DuckDB one is the right call — it resolves partitioned append, upsert, JSON and cross-process safety structurally instead of one at a time, and catalog-wide coherent snapshots are a capability the old shape couldn't have offered.

Putting the key on `DatasetSpec.primary_key` rather than on the write call is better than passing `key_columns` per write: the key is a property of the table, and it avoids a parameter that's conditionally required on one mode. Test coverage is solid too, including `test_upsert_updates_inserts_and_moves_partition_keys`, which is the case most likely to silently duplicate.

Three things I think should change.

---

### 1. `write(name, [], mode="replace")` is now a no-op — regression

`DatasetSecondaryAdapterDuckLake.write`, lines 176-185:

```python
if not normalized_rows:
    return spec                      # ← returns here

self._create_incoming_table(con, spec)
self._insert_incoming_rows(con, spec, normalized_rows)
target = self._qualified_table(namespace, name)
columns = ", ".join(self._ident(column.name) for column in spec.columns)

if mode == "replace":
    con.execute(f"DELETE FROM {target}")
```

The empty-rows early return sits before the `DELETE`, so replacing with an empty list silently does nothing.

The deleted adapter truncated in this case — its `rmtree` ran *before* the `if rows:` guard:

```python
if mode == "replace" and data_dir.exists():
    shutil.rmtree(data_dir)
data_dir.mkdir(parents=True, exist_ok=True)
if rows:
    self._copy_rows(info, rows, data_dir)
```

"Empty this dataset" is a legitimate call and it now fails silently. No test covers it — `test_replace_overwrites_rows` uses non-empty rows. Suggest moving the early return below the mode dispatch (or special-casing `replace`), plus a generic test for it.

### 2. `DatasetSnapshotNotFoundError` doubles as the optimistic-concurrency conflict

Lines 168-170:

```python
current_snapshot = self._current_snapshot(con)
if snapshot_id is not None and snapshot_id != current_snapshot:
    raise DatasetSnapshotNotFoundError(snapshot_id)
```

The snapshot usually *does* exist here — it's stale, not missing. A caller doing compare-and-swap can't distinguish "you passed a snapshot that never existed" (a bug) from "someone committed before you" (retry). Catching a not-found error to mean conflict makes retry loops read wrong and swallows genuine bugs.

Suggest a distinct `DatasetSnapshotConflictError`, and keeping `DatasetSnapshotNotFoundError` for `query(snapshot_id=...)` where it really is a lookup miss.

### 3. The snapshot `write()` returns can belong to another writer

In `_retry_write_transaction`:

```python
con.execute("COMMIT")
return value, self._current_snapshot(con)
```

The current snapshot is read *after* the commit, so under concurrency another writer can commit in between and the returned `DatasetInfo.snapshot_id` is theirs, not the one this write produced. It's benign for control flow, but wrong for anything that records the returned snapshot as provenance ("these rows landed in snapshot N"), which is exactly what a versioned store invites callers to do.

If DuckLake exposes the committed snapshot from the transaction itself, taking it from there would be exact. Otherwise it's worth a docstring saying the returned id is "the catalog snapshot observed after commit", not "the snapshot this write created".

---

### Two notes, not change requests

**`primary_key` is advisory.** The ADR is explicit that DuckLake has no PK or unique constraints, so it only drives the `MERGE` match — `mode="append"` into a keyed dataset still duplicates silently. That's the right trade, but it's load-bearing for callers and currently only stated in the ADR. Worth putting in the `DatasetSpec.primary_key` docstring, since that's where someone will read it.

**Snapshots are catalog-wide, so CAS is catalog-wide.** Any write to any dataset bumps the counter, so passing `snapshot_id=` as a concurrency token gets more contended the more datasets share a catalog. Not a defect — a consequence of coherent snapshots, which are worth it — but callers with several datasets and concurrent writers should be told to rely on `upsert` plus the retry policy rather than explicit CAS.

---

### For the MinIO/S3 follow-up

Flagging early since object-storage-backed data paths look imminent. `_prepare_local_paths` already skips the mkdir for a URI data path, but `_connect()` only loads `ducklake`, and `DatasetAdapterDuckLakeConfiguration` carries no S3 fields — so a remote `data_path` can't authenticate yet.

What that needs:

- `INSTALL httpfs; LOAD httpfs;` alongside ducklake, same fallback pattern
- a per-connection `CREATE OR REPLACE SECRET (TYPE s3, ...)` — a `PERSISTENT` secret writes to a local secret store, which isn't right for a container
- config fields for key/secret/endpoint/region

Three MinIO-specific details that are easy to get wrong:

- `URL_STYLE 'path'` is mandatory; without it DuckDB builds virtual-host URLs (`http://<bucket>.minio:9000/...`) that MinIO won't serve
- `ENDPOINT` wants `host:port` with **no scheme** — the existing `object_storage` config uses `endpoint_url: "http://minio:9000"`, so reusing that value directly will fail
- `USE_SSL false` for an in-cluster endpoint

Two more worth handling in the same change:

- **Reject a remote `data_path` paired with a `sqlite:` catalog.** Shared data files with a per-container catalog means each process sees its own set of tables while writing into the same warehouse. It produces no error at all, just divergence, so a config validator is worth more here than documentation.
- **`_is_retriable` needs transient HTTP.** It currently matches `TransactionException` plus `database is locked` / `transaction conflict` / `serialization conflict` / `failed to commit`. Object storage adds timeouts, 503s and connection resets, which would surface as hard failures.

Finally, worth reconsidering connect-per-operation once the catalog and warehouse are both remote. Today every call is a fresh connection, extension load, secret creation and `ATTACH`. That's cheap against local files and several round trips against Postgres + S3.

Dr0p42:
Re-reviewed through `61d5aa1`. All three change requests are resolved, and two of them went further than I asked:

- **Empty `replace`** — `test_replace_with_empty_rows_clears_dataset` in the generic suite, so every adapter inherits it. The empty-rows branch also keeps the unknown-mode `ValueError`, which closes a hole I hadn't spotted: an unknown mode with an empty batch used to return silently.
- **Conflict error** — `DatasetSnapshotConflictError` carries both `expected_snapshot_id` and `current_snapshot_id`. I'd only asked for a distinct type; carrying both means a retry loop can log the gap rather than re-query for it.
- **Snapshot returned by `write()`** — `last_committed_snapshot()` is exactly right, and the `None` fallback for transactions that create no snapshot is the case I'd have missed. `test_write_returns_the_snapshot_committed_by_its_connection` and `test_noop_write_returns_the_current_snapshot` cover both paths.

The two documentation notes landed where they'll actually be read — the advisory-key caveat on `DatasetSpec.primary_key` itself, and the catalog-wide CAS semantics in the `write` docstring with the "prefer idempotent upserts and adapter retries" steer.

The S3 work from `54d7e00` is good. Accepting a scheme-ful endpoint and stripping it is better than what I suggested (documenting that it must be scheme-less), and the `_configure_object_store` docstring naming the silent-inlining failure mode — a batch small enough to be inlined in the catalog appears to succeed while never reaching the store — is the kind of thing that saves someone a day.

Five things still outstanding. None blocks the merge in my view, but the first is the one most likely to burn someone.

---

### 1. A scheme-less `s3_endpoint` silently selects TLS

In `_S3Settings.sql()`, `use_ssl` is only inferred inside the scheme-matching loop:

```python
use_ssl = self.use_ssl                          # None
for scheme, ssl in (("https://", True), ("http://", False)):
    if endpoint.startswith(scheme):             # "minio:9000" matches neither
        ...
parts.append(f"USE_SSL {'true' if (use_ssl is not False) else 'false'}")
```

So `s3_endpoint: "minio:9000"` — the form DuckDB's own docs use for `ENDPOINT` — yields `USE_SSL true` and fails against a plain-HTTP MinIO. Only the `http://` and `https://` forms are tested.

It's the mirror image of the trap the scheme-stripping fixed: whichever of the two forms someone writes, one of them is wrong, and neither reports why. Defaulting `USE_SSL` to `false` when a custom endpoint is set and nothing else determines it would match the reasoning already in the `URL_STYLE` comment two lines above — a custom endpoint is almost never AWS. Alternatively, require the scheme and raise without it.

### 2. Three of the seven advertised schemes cannot authenticate

`OBJECT_STORE_SCHEMES` admits `gcs://`, `gs://`, `r2://`, `azure://` and `abfss://`, but `_S3Settings.sql()` always emits `TYPE s3`. DuckDB wants `TYPE gcs` for GCS, `TYPE r2` with an `ACCOUNT_ID` for R2, and `TYPE azure` for Azure.

The constructor guard only rejects credentials on a *non*-object-store path, so `data_path="gs://bucket/…"` plus S3 credentials passes and builds a secret DuckDB will never apply — landing back on the silent-failure path the `_configure_object_store` docstring warns about. Either narrow the tuple to the S3-compatible schemes, or let the secret `TYPE` follow the scheme.

### 3. `_is_retriable` has no transient-HTTP markers

It still matches `TransactionException` plus `database is locked` / `transaction conflict` / `serialization conflict` / `failed to commit`. That was the complete set when the warehouse was a local disk. Now that every write goes to object storage, timeouts, 503s and connection resets are ordinary events, and they currently surface as hard failures straight through a retry policy built for exactly this.

### 4. A remote `data_path` with a `sqlite:` catalog is still only documented

The config docstring shows the correct pairing (`postgres:` catalog with an `s3://` data path), but nothing rejects the wrong one. Shared data files with a per-container SQLite catalog means each process sees its own set of tables while writing into the same warehouse — no error, just divergence, and the symptom (tables that exist for one container and not another) points nowhere near the cause. A validator is worth more than the docstring here.

### 5. The warehouse and catalog sit inside the object-storage root, in the shipped default

`config.yaml`:

```yaml
object_storage: base_path: "storage/datastore"
dataset:        data_path: "storage/datastore/datasets/"
                catalog:   "sqlite:storage/datastore/datasets.sqlite"
```

So anything enumerating the object storage root sees the Parquet files and the catalog database as if they were user objects — and can hand them to a caller as content. The MinIO example nests identically (`s3://abi/abi/datastore/datasets/` against the `abi` bucket).

That's harmless while nothing walks the tree, but ingestion pipelines that take "a storage prefix" as input will walk straight into the warehouse. Given both defaults ship together, a sibling path (`storage/datasets/`) or a dedicated bucket would keep the two from overlapping by construction.

Dr0p42:
`52415e7` resolves the endpoint issue, and two of my other points were already answered in `61d5aa1` — I checked the code and not the docs, so one correction below.

**Endpoint transport.** Requiring an explicit choice is the safer of the two options I offered; guessing `false` for a custom endpoint would have quietly disabled TLS for someone pointing at a real HTTPS-only store. Enforcing it in both `_S3Settings.__post_init__` and the config validator is better than either alone — the config layer fails at load with a message naming the fix, the dataclass catches programmatic construction. Two things I hadn't raised also got handled: a non-HTTP scheme like `s3://minio:9000` is now rejected rather than half-parsed, and scheme matching is case-insensitive.

**Correction — withdrawing my transport-retry request.** The ADR already answers it, and the answer is better than my suggestion:

> Object-store timeouts and connection failures are not transaction retries: replay after an ambiguous commit could duplicate an append. Safe transport retries require an idempotency or commit-reconciliation contract.

That's right and I was wrong. A timeout can arrive after the metadata commit succeeded, and `append` carries no key, so replaying it duplicates rows. Folding transport failures into `_is_retriable` would have turned a visible failure into silent duplication — precisely the class of bug the rest of this design works to avoid. Surfacing them to the caller is correct until the port has an idempotency contract.

**Also already answered — the SQLite-plus-remote-path pairing.** Both the ADR and `AGENTS.md` state that a remote data path may be deliberately paired with SQLite for a single-process runtime, and that replicas need one durable shared catalog. That's a legitimate configuration, so a hard validator would be wrong. My "worth more than the docstring" line was posted after you'd already written the rationale.

Two things from that comment do still stand:

- **`OBJECT_STORE_SCHEMES` vs `TYPE s3`.** The tuple still admits `gcs://`, `gs://`, `r2://`, `azure://` and `abfss://` while `_S3Settings.sql()` always emits `TYPE s3`. A `gs://` data path with credentials passes the constructor guard and builds a secret DuckDB won't apply, which lands on the silent-inlining path rather than an error. Narrowing the tuple would be the smaller fix if the other stores aren't a near-term target.
- **The default paths overlap.** `object_storage.base_path: storage/datastore` still contains both `dataset.data_path` and the catalog file, so anything enumerating the object-storage root sees Parquet files and the catalog database as user objects. A sibling path would separate them by construction.

Neither blocks the merge. Nice work on the turnaround.

Dr0p42:
`37ffd6b` closes both remaining points. Nothing outstanding from my side.

**Schemes.** Narrowing to `s3://`/`s3a://` rather than teaching `_S3Settings` the other secret types is the right scope call — an unsupported scheme now fails at config load naming the supported set, instead of building a secret DuckDB silently ignores. Rejected in the adapter constructor *and* the config validator, parametrized over all five, matching how the endpoint check is enforced.

It also incidentally tightened `_prepare_local_paths`: deciding the local branch positively with `not _is_s3_data_path(...)` rather than negatively on `"://" not in ...` means an unrecognised scheme can no longer skip the mkdir and fail later at write time.

**Path isolation.** `storage/datasets/` and `storage/datasets.sqlite` as siblings of `storage/datastore` is the version I was hoping for — the separation is structural, so nothing has to remember to exclude a prefix. Carried through both configs, both CLI templates, the engine defaults, the Docker backup note and the ADR, with `test_dataset_defaults_do_not_overlap_object_storage_namespace` to keep them apart.

Full ledger across the four rounds: eight items fixed, two answered by design decisions recorded in the ADR — the transport-retry one being a case where the recorded reasoning was better than my request. The consistent pattern of enforcing at both the config and adapter layers, and putting behavioural tests in the generic suite so future adapters inherit them, is what made this quick to re-review.

LGTM.

Issue #15, opened 2023-10-23
As a chat user, I would like to get the list of people that liked or commented my last x posts


jravenel:
Objective: I want to be able to identify if the list of people match my ICP = Marketing Qualitfied Leads. 
I should be able to give a prompt as input to identify the ICP it should return a graph with total number of people for each ICP category, a markdown table with all the data, and a CSV file to be downloaded and further analyzed.

PR #1258, opened 2026-09-09
feat(core): add portable document store service

### LINKED ISSUE #1257: Document store service: portable per-module storage for modules and apps
## Summary

Add a **document store service** to `naas-abi-core`: a narrow, portable
collection/document API that lets any module or app declare its own storage and
query it, without knowing or caring which backend serves it.

The port is deliberately small — CRUD plus structured filters, no joins, no
cross-document transactions, no SQL escape hatch — because portability is
inversely proportional to port surface. A module written against it runs on
SQLite locally and PostgreSQL in production with no code change and no
configuration awareness.

## Motivation

Today a module that needs to store application state has three bad options.

**1. Use the dataset service.** It is designed for large analytical data as
Parquet, with catalog-wide snapshots and time travel. Right for a research
corpus; wrong for small mutable app state. Every write creates a snapshot, there
are no row-level updates, and no transactional claim.

**2. Use the key-value service.** Its port is `get` / `set` /
`set_if_not_exists` / `delete` / `delete_if_value_matches` / `exists`. There is
**no enumeration** — no `keys`, `scan` or `list` — so "find all pending records"
is inexpressible. Anything needing a query has to hand-roll an index key plus a
lock, i.e. rebuild a query engine out of primitives that deliberately lack one.

**3. Connect to PostgreSQL directly.** This is what Nexus does. It works, but the
DSN is hand-configured and the database must already exist — which produced a
real outage: `003-create-coder-db.sql` and `004-create-forgejo-db.sql` were added
to `.deploy/docker/postgres/initdb/` after the Postgres volume existed. Postgres
only runs `initdb` scripts on an **empty** data directory, so they silently never
ran. Coder crash-looped 47 times with `pq: database "coder" does not exist`, and
because `dagster` declares `depends_on: coding-init: service_completed_successfully`,
a host reboot would have left the whole orchestration layer down indefinitely.
Nothing in `docker compose ps` showed a problem.

A module should be able to say "I need somewhere to keep records" and get it,
idempotently, on every boot — the same contract `ensure_datasets()` already
provides for datasets.

## Considered and rejected

**A PostgreSQL adapter behind the existing dataset port.** The dataset port is
shaped around analytical semantics (snapshots, time travel, upsert-by-rewrite).
Adding row updates, transactions and `SKIP LOCKED` would widen it until it is
the union of two incompatible models, where every adapter implements half and
raises on the rest. Callers would have to know which half they are on, so the
shared interface would cost clarity and buy nothing.

**A provisioning service that hands back a DSN or connection.** Attractive
because it preserves full `psycopg` power and would fix the initdb class of bug.
Rejected because it makes modules *deployment-coupled*: a module that receives a
PostgreSQL DSN only runs where PostgreSQL exists, regardless of which file the
SQL sits in. That breaks the portability guarantee this service exists to
provide. It remains a reasonable future addition as an explicitly-declared
escape hatch, but it must not be the default and must not be how ordinary
modules store state.

**An ORM in core.** Worse lock-in than backend coupling, not better. Every module
would be pinned to core's SQLAlchemy version and welded to its session, identity
map and lazy-loading model. Databases are easier to replace than ORMs. Note the
existing dependency split already reflects this: `naas-abi-core` depends on
`psycopg[binary,pool]` (driver + pool, no ORM) while `naas-abi` depends on
`sqlalchemy` and `alembic` one layer up, where Nexus uses them. Keep that split.

**MongoDB as a required dependency.** Adds an eighth datastore and buys nothing
that PostgreSQL `JSONB` does not already provide. A Mongo adapter is welcome;
a Mongo requirement is not.

### Value types

```python
Value = str | int | float | bool | None | datetime | bytes | list["Value"] | dict[str, "Value"]
```

`datetime` must be timezone-aware; naive datetimes are the richest source of
cross-backend bugs and should be rejected at the service boundary. `Decimal` is
deliberately excluded for now — Mongo has `Decimal128`, PostgreSQL has `numeric`,
SQLite has neither, so it would be tag-encoded and lossy to compare. Add it when
something needs money.

### Specs

```python
Operator = Literal["eq", "ne", "lt", "lte", "gt", "gte", "in", "nin", "contains", "exists"]
Predicate = tuple[str, Operator, Value]          # ("runs", "gte", 10)

class FieldSpec(BaseModel):
    name: str
    type: Literal["string", "int", "float", "bool", "datetime", "bytes", "json"]
    indexed: bool = False        # optimisation hint; never limits what is queryable
    unique: bool = False

class CollectionSpec(BaseModel):
    """Declared fields are not a schema in the SQL sense: undeclared fields are
    still storable and queryable. Declaring buys correct typing on read and
    correct ordering on sort, which JSON alone cannot give you — a stored date is
    indistinguishable from a string that looks like one."""
    name: str
    fields: tuple[FieldSpec, ...] = ()
    unique_together: tuple[tuple[str, ...], ...] = ()
```

### Returned shapes

```python
@dataclass(frozen=True)
class Document:
    id: str
    data: dict[str, Value]
    created_at: datetime
    updated_at: datetime
    version: int          # compare-and-swap counter, NOT a schema version

@dataclass(frozen=True)
class Page:
    items: list[Document]
    cursor: str | None    # opaque keyset cursor; None means last page
```

Cursor paging rather than offset: offset degrades badly on Mongo at depth and is
not expressible on DynamoDB. The cursor encodes the last `(sort_key, id)` pair,
which is stable under concurrent inserts.

### The port (implemented by adapters)

```python
class IDocumentAdapter(Protocol):
    def ensure_collection(self, namespace: str, spec: CollectionSpec) -> None: ...
    def drop_collection(self, namespace: str, collection: str) -> None: ...
    def collections(self, namespace: str) -> list[str]: ...

    def put(self, namespace: str, collection: str, id: str,
            data: dict[str, Value], if_version: int | None) -> Document: ...
    def get(self, namespace: str, collection: str, id: str) -> Document: ...
    def delete(self, namespace: str, collection: str, id: str,
               if_version: int | None) -> None: ...

    def find(self, namespace: str, collection: str, where: Sequence[Predicate],
             order_by: tuple[str, Literal["asc", "desc"]] | None,
             limit: int, cursor: str | None) -> Page: ...
    def count(self, namespace: str, collection: str,
              where: Sequence[Predicate]) -> int: ...
```

### The service (called by modules)

```python
class DocumentService(ServiceBase):
    """Documents in collections, scoped to a namespace.

    One document is the unit of atomicity. There are no joins and no
    multi-document transactions: shape a document so that what changes together
    lives together, and use `if_version` for safe read-modify-write.
    """

    def __init__(self, adapter: IDocumentAdapter, namespace: str): ...

    # collections
    def ensure_collection(self, spec: CollectionSpec) -> None: ...   # idempotent, every boot
    def drop_collection(self, collection: str) -> None: ...
    def collections(self) -> list[str]: ...

    # single document
    def put(self, collection: str, id: str, data: dict[str, Value],
            *, if_version: int | None = None) -> Document: ...
    def get(self, collection: str, id: str) -> Document: ...          # raises DocumentNotFound
    def find_one(self, collection: str, where: Sequence[Predicate]) -> Document | None: ...
    def exists(self, collection: str, id: str) -> bool: ...
    def delete(self, collection: str, id: str, *, if_version: int | None = None) -> None: ...

    # queries
    def find(self, collection: str, *, where: Sequence[Predicate] = (),
             order_by: tuple[str, Literal["asc", "desc"]] | None = None,
             limit: int = 100, cursor: str | None = None) -> Page: ...
    def iterate(self, collection: str, *, where=(), order_by=None,
                batch: int = 500) -> Iterator[Document]: ...          # pages internally
    def count(self, collection: str, where: Sequence[Predicate] = ()) -> int: ...

    # bulk — per-document atomicity, NOT a transaction
    def put_many(self, collection: str,
                 items: Mapping[str, dict[str, Value]]) -> list[Document]: ...
    def delete_many(self, collection: str, where: Sequence[Predicate]) -> int: ...
```

Exceptions: `DocumentNotFound`, `VersionConflict`, `UniqueViolation`,
`CollectionNotFound`.

`namespace` is per-module, so modules cannot read each other's collections.

### How a module consumes it

The service is **never** referenced from a module's domain. The module defines
its own port; an adapter binds it to `DocumentService`:

```python
# requests/interfaces.py — the domain's port
class RequestStore(Protocol):
    def save(self, request: RunRequest) -> None: ...
    def pending(self) -> list[RunRequest]: ...

# requests/adapters/secondary/DocumentRequestStore.py — technology lives here
class DocumentRequestStore:
    def __init__(self, documents: DocumentService):
        self._docs = documents

    def pending(self) -> list[RunRequest]:
        page = self._docs.find("run_requests",
                               where=[("status", "eq", "pending")],
                               order_by=("requested_at", "asc"))
        return [_to_request(d) for d in page.items]
```

`requests/domain.py` continues to import nothing but `interfaces`.

### PostgreSQL

One table per module namespace/schema; no DDL per collection.

```sql
CREATE TABLE documents (
    collection  text        NOT NULL,
    id          text        NOT NULL,
    data        jsonb       NOT NULL,
    created_at  timestamptz NOT NULL DEFAULT now(),
    updated_at  timestamptz NOT NULL DEFAULT now(),
    version     bigint      NOT NULL DEFAULT 1,
    PRIMARY KEY (collection, id)
);
CREATE INDEX ON documents USING gin (data jsonb_path_ops);
```

`find` translation:

```sql
SELECT id, data, created_at, updated_at, version
FROM documents
WHERE collection = $1
  AND data @> '{"active": true}'::jsonb        -- eq  → containment, uses GIN
  AND (data->>'runs')::numeric >= $2           -- range → cast, needs expression index
ORDER BY data->>'started_at' DESC
LIMIT 50;
```

Compare-and-swap, which is what makes `version` work without transactions:

```sql
UPDATE documents
SET data = $3, version = version + 1, updated_at = now()
WHERE collection = $1 AND id = $2 AND version = $4
RETURNING version;          -- zero rows → VersionConflict
```

Keyset paging:

```sql
  AND (data->>'started_at', id) < ($cursor_ts, $cursor_id)
ORDER BY data->>'started_at' DESC, id DESC
```

Uniqueness:

```sql
CREATE UNIQUE INDEX ON documents ((data->>'email')) WHERE collection = 'users';
```

Two traps the adapter must handle:

- **`ORDER BY data->>'x'` sorts as text.** `"9" > "10"` lexically, so numeric
  sorts are silently wrong without a cast driven by the declared field type. ISO-8601
  UTC dates happen to sort correctly as text; numbers do not.
- **`eq` and range predicates have different index behaviour.** Containment rides
  the GIN index; `>=` on an extracted field needs an expression index or it is a
  sequential scan. This is what `FieldSpec.indexed` is for.

If JSONB extraction later proves too slow, the adapter can promote declared
fields to generated columns with real types, indexes and planner statistics —
entirely invisibly, because the port is narrow:

```sql
ALTER TABLE documents ADD COLUMN started_at timestamptz
  GENERATED ALWAYS AS ((data->>'started_at')::timestamptz) STORED;
```

### SQLite

Same shape, `json_extract(data, '$.field')` in place of `->>`. Expression indexes
supported since 3.9. This is the adapter that removes PostgreSQL from the local
development loop entirely — a module and its tests run against a file instead of
requiring the whole compose stack.

### MongoDB

Native fit; needs almost none of the marshalling because BSON already carries the
types.

### Type round-trip requirements

| value | PostgreSQL JSONB | SQLite | MongoDB |
|---|---|---|---|
| `True` | native JSON `true` | `INTEGER 1` → decoded back to `True` | native `Boolean` |
| `datetime` | ISO string, typed by `FieldSpec` | `TEXT`, typed by `FieldSpec` | native BSON `Date` |
| `bytes` | base64 + type tag | `BLOB` | native `BinData` |
| `None` | JSON `null` | `NULL` | `null` |

For values a backend cannot represent natively and that are not covered by a
declared field type, the adapter writes a tagged envelope
(`{"$t": "bytes", "$v": "<base64>"}`) and unwraps it on read. This is invisible to
modules. Note that tagging breaks `@>` containment on that field, which is why
declared types are preferred for anything queryable.

## Uniqueness

Single-field via `FieldSpec.unique`, compound via
`CollectionSpec.unique_together`. Supported on PostgreSQL, SQLite and MongoDB.
Not supported on DynamoDB, which has no uniqueness beyond the partition key —
consistent with DynamoDB already being out of scope (see Non-goals).

**NULL semantics must be decided explicitly**, because backends genuinely differ:

- PostgreSQL / SQLite: NULLs are *distinct*, so any number of documents may omit
  the field
- MongoDB: a missing field indexes as `null`, so with a plain unique index **only
  one** document may omit it, unless the index is declared `sparse`

Proposal: **unique indexes are always sparse.** PostgreSQL and SQLite get this for
free; the Mongo adapter must pass `sparse: true`. Leaving this unspecified is the
one thing that would silently break portability.

`ensure_collection` must **fail loudly** when asked to add a unique constraint to
a collection whose existing data violates it, rather than skipping silently.

## Migrations

Documents remove **structural DDL migrations** — adding a field is just writing a
document that has it. No `ALTER TABLE`, no lock on a large table, no deploy
ordering between code and schema.

They do **not** remove **data migrations**, and they make them silent. Renaming
`chunk_size` to `chunkSize`, or narrowing `status` to a fixed set, leaves every
existing document in the old shape. SQL forces that to be confronted at migration
time; documents defer it into the read path, where it usually shows up as
`doc.get("chunkSize") or doc.get("chunk_size")` spreading through the codebase.

Recommended discipline:

- A `_schema: int` field on the document, with **one** upcast function at the
  adapter boundary that migrates old versions to the current shape on read. Keep
  this strictly separate from `Document.version`, which is the compare-and-swap
  counter — conflating them would be a subtle and nasty bug.
- Idempotent, resumable backfills using `if_version` when an old shape must
  actually be retired.

Index and constraint changes remain migrations regardless of storage model.

## Generic secondary-adapter test suite

Following the existing pattern
(`services/dataset/tests/dataset__secondary_adapter__generic_test.py`), a new
adapter is considered valid when it passes
`document__secondary_adapter__generic_test.py`. This is what makes "works on any
backend" a fact rather than an aspiration.

Required cases at minimum:

```python
ROUND_TRIP_VALUES = [
    True, False, 0, -1, 1.5, 0.0, "", "café", None, [], {}, [1, [2]],
    {"nested": {"deep": True}}, datetime(2026, 1, 1, tzinfo=UTC), b"\x00\xff",
]

def test_every_value_round_trips(adapter):
    """True comes back True, not 1. A datetime comes back a datetime, not a string."""

def test_ordering_is_by_declared_type_not_lexical(adapter):
    """9 sorts before 10 — the sharpest edge in the design."""

def test_compare_and_swap_rejects_a_stale_version(adapter): ...
def test_cursor_paging_is_stable_under_concurrent_inserts(adapter): ...
def test_duplicate_violating_unique_raises_UniqueViolation(adapter): ...
def test_two_documents_omitting_an_optional_unique_field_both_succeed(adapter): ...
def test_naive_datetime_is_rejected(adapter): ...
def test_ensure_collection_is_idempotent(adapter): ...
def test_namespaces_are_isolated(adapter): ...
```

## Non-goals

Each of these is excluded for a reason, not an oversight.

- **Joins.** Every backend can implement one, but at wildly different cost —
  PostgreSQL has a planner, Mongo `$lookup` is effectively a nested loop, DynamoDB
  cannot do it at all and would need N+1 round trips. A port that implies uniform
  cost while hiding a 1000× spread is a promise the caller cannot verify. It also
  requires join keys, join types, result projection and filter composition —
  at which point the port has become a query language, i.e. a homegrown ORM.
- **Cross-document transactions.** One document is the unit of atomicity.
  DynamoDB and Mongo cannot honour more than that in the general case. Shape
  aggregates so what changes together lives together.
- **Foreign keys / referential integrity.** Requires a join to check and a
  transaction to maintain. Dangling references are the domain's problem.
- **A raw `query(sql)` escape hatch.** It would make modules non-portable the day
  it shipped, and ports can be widened but never narrowed.
- **DynamoDB as a target.** Supporting it would force `find` down to
  key-plus-declared-index access, costing every module real expressiveness for a
  backend not currently in use. An adapter can be added later if someone accepts
  those constraints for their own deployment.

Ports are easy to widen and impossible to narrow: ship the smallest surface that
serves a real module, and let the second and third modules widen it with
evidence.

## Relationship to existing services

- **dataset** — unchanged. Large analytical data, columnar, Parquet, snapshots,
  time travel, joins. When document-shaped state needs analytical querying, the
  answer is to *project* it into a dataset, the same way `phases_v2` already
  projects datasets into the triple store and vector store.
- **keyvalue** — unchanged. Remains the right tool for locks and ephemeral
  values; `set_if_not_exists` already provides a distributed-lock primitive that
  complements this service.
- **DocumentService** — new. Module and application state.

## Open decisions

1. Namespace isolation on PostgreSQL: schema-per-module (`CREATE SCHEMA <module>`)
   versus a `namespace` column. Schema-per-module gives cleaner teardown and
   permissions; a column avoids DDL.
2. Whether `find` should support projection (returning a subset of fields) or
   always return whole documents.
3. Whether `contains` applies to arrays only, or also to substring matching on
   strings — the latter has very different index behaviour per backend.
4. Whether `ensure_collection` may *drop* an index that is no longer declared, or
   only ever add.

## Acceptance criteria

- [ ] `DocumentPort`, `DocumentService`, `DocumentFactory` in `naas-abi-core`
- [ ] PostgreSQL adapter (`psycopg`, already a core dependency — no new dependency)
- [ ] SQLite adapter, enabling module development with no external services
- [ ] `document__secondary_adapter__generic_test.py`, passing for both
- [ ] Configurable through `config.yaml` `services:` like every other service
- [ ] Declarable in `ModuleDependencies.services`, so a deployment that cannot
      satisfy it fails at load rather than at runtime
- [ ] Documented: the atomicity boundary, the absence of joins, and the
      migration discipline

Dr0p42:
Modules can store mutable application state through one collection/document API on SQLite and PostgreSQL. Declaring `DocumentService` loads the configured backend before module initialization, and the module proxy supplies a service scoped to the module's full name.

Closes #1257.

## Changes

- Add the document port, service, factory, and SQLite/psycopg adapters without new runtime dependencies.
- Support CRUD, structured filters, typed sorting, cursor paging, version-checked writes/deletes, and sparse single-field/compound uniqueness.
- Preserve nested values, including aware datetimes and bytes; reject values that cannot be represented consistently on both backends.
- Provision storage idempotently and persist additive collection declarations. Adding an invalid uniqueness constraint fails transactionally.
- Wire engine dependency loading, module scoping with preserved service wiring, YAML configuration, and generated project templates. Remote configurations explicitly use PostgreSQL.
- Use concurrent SQLite file transactions and a bounded PostgreSQL connection pool, with explicit adapter cleanup. Match SQLite equality queries to expression indexes and repair outdated indexes transactionally.
- Materialize iterable filters before validation and retain them across pages, cap page/batch sizes at 1,000, and canonicalize AND-predicate order for cursors. Translate backend failures to `DocumentStorageError`.
- Reuse scoped module handles with access checks, require explicit standalone namespaces, cache parsed specs without skipping catalog locks, and track PostgreSQL index definitions for transactional upgrades.
- Align numeric equality, uniqueness, ranges, and paging on JSON decimal values. Bound optional PostgreSQL text index keys so long strings/bytes remain writable, and repair previous index definitions transactionally.
- Add a dedicated document-test CI job on Python 3.11/3.12 with PostgreSQL 17.
- Document query semantics, per-document atomicity, migration discipline, and deployment choices in the service guide and [ADR](docs/adr/20260909_portable-document-service.md).

## Validation

- **253 tests passed on each of Python 3.11 and 3.12**, using SQLite 3.53.3/3.46.0 respectively and a real PostgreSQL 17.11 instance. The new `make test-document-core` CI target covers both adapter contracts, direct port/codec/compiler tests, lifecycle, query plans, engine wiring, and remote configurations.
- Regression coverage includes quoted literal keys and uniqueness on older SQLite, concurrent snapshots, transactional index repair, pool reuse/rollback/close, pagination through every value kind, and large-float reads/updates/paging after JSONB number normalization.
- Additional regressions cover iterator-safe bulk deletion, cursor predicate reordering, page bounds, PostgreSQL stale-index repair/rollback, lock/pool timeout translation, cross-adapter catalog changes, and cached-proxy access checks.
- Regressions also cover mixed large floats/integers (including negative values and nested equality), old numeric-index migration/rollback, long incompressible ASCII/Unicode strings and bytes, and optional PostgreSQL index upgrades.
- The full `make check` commit hook passed: core/CLI lint and mypy (541 core files and 80 CLI files), Bandit, marketplace lint/type checks, and the Nexus production build.
- The initial broader engine/template run also exposed three existing engine-configuration failures, each reproduced against baseline code: `test_load_configuration_allows_disabling_api_reload`, `test_load_configuration_fails_if_configured_dotenv_file_is_missing`, and `test_from_yaml_content_loads_opencode_provider_configuration`.
- Built and inspected the core wheel during initial implementation.

The PostgreSQL test fixture uses `DOCUMENT_TEST_POSTGRES_DSN`, creates an isolated schema, and removes it afterward. It explicitly skips if the variable is absent.

## Operational choices

Whole documents are returned; `contains` means exact array membership; declarations only add indexes/constraints. Dictionary key order is unspecified; lists preserve order. SQLite equality and `in` can use expression indexes; exclusions and array membership may scan. PostgreSQL uses namespace columns in a dedicated service schema within an existing database. Credentials need schema/table/index creation permission. SQLite is the default for local development.

Numbers compare by their JSON decimal value, rather than the exact binary expansion of Python floats. Integral floats can normalize to integers within signed 64 bits; expanded numbers outside that range are restored to floats so retrieved documents remain valid portable values.

Optional PostgreSQL B-tree text/bytes keys use a bounded prefix; queries and cursors compare complete values, so text sorting may require an explicit sort.

There are no joins or multi-document transactions. Bulk helpers may commit a prefix before an error, and cursors do not provide snapshot isolation. Module scoping is an application boundary, not a sandbox for hostile Python code. Timeout defaults and the absence of automatic write retries are documented in the ADR.

Incompatible declarations deliberately fail at boot with collection/document context; migrate the data before enabling those declarations. Supplied delete versions must match an existing document, including zero. PostgreSQL index comments are owned by the adapter to track definition fingerprints, and catalog locks remain per operation for correctness across processes.


Dr0p42:
## Review

xhigh-effort multi-angle review (verified against the worktree with live `pytest` runs). 15 findings, ranked most severe first.

### Correctness bugs

**1. SQLite field-name JSON-path escaping is broken — silently corrupts queries and unique constraints**
`libs/naas-abi-core/naas_abi_core/services/document/adapters/secondary/document_sql.py:63`
`field()` builds SQLite JSON paths as `"$." + json.dumps(key)`, but SQLite's JSON path parser doesn't handle the `\"` escaping `json.dumps()` produces for a field name containing a literal `"`. **Empirically confirmed**: the PR's own adversarial test (`test_names_and_predicate_values_are_not_sql`, field name `a.b'"$`) fails — `find()` returns zero rows for that field, and worse, `FieldSpec(unique=True)` on such a field name has its uniqueness silently unenforced (two dupes insert with no error). Postgres is unaffected (plain literal key), so SQLite/Postgres diverge on the same field name — undermining the PR's core portability promise. Local run: **1 failed, 73 passed, 51 skipped.**

**2. `DocumentService._for_namespace()` drops the engine's `services` wiring**
`libs/naas-abi-core/naas_abi_core/services/document/DocumentService.py:37`
Constructs a new instance via the plain constructor, resetting `ServiceBase._services` to `None`. `EngineProxy.ServicesProxy.document` calls this on every module access, so every module-scoped `DocumentService` handle is permanently unwired. Dormant today (nothing calls `self.services` yet), but the first use that does (matching every sibling service's pattern) raises `AssertionError: Services are not wired` in production, with no test currently asserting `services_wired` on a per-module instance.

**3. `config.remote.yaml` has no `document:` entry — remote deployments silently fall back to local SQLite**
`config.remote.yaml:81` (and the same gap in `libs/naas-abi-cli/.../templates/project/config.remote.yaml`)
Every other stateful backend is routed off-box in that file (S3, TDB2, Qdrant, RabbitMQ); `document` is missing entirely, so Pydantic's default silently applies: ephemeral local SQLite at `storage/documents.sqlite`. Any module or scaffolded project deployed remotely gets non-shared, non-persistent storage with no config error to flag it.

**4. Declared SQLite indexes don't accelerate the queries they're declared for**
`libs/naas-abi-core/naas_abi_core/services/document/adapters/secondary/document_sql.py:118`
`predicates()`'s eq/ne/in/nin/contains branches build a different SQL expression than the one `ensure_collection()` actually indexes (via `sort_parts()`). A field marked `indexed=True` gets no acceleration on equality/membership lookups — only `ORDER BY`/range benefits. On a large SQLite collection, `find(where=[("status","eq","active")])` does a full table scan despite the declared index. (Postgres is fine — it has a separate GIN index.)

**5. `_for_namespace()` has no access control — any handle can mint a handle for any other module's namespace**
`libs/naas-abi-core/naas_abi_core/services/document/DocumentService.py:37`
The per-module ACL is enforced once, at `EngineProxy.ServicesProxy.document`'s access check, never again on the returned object. `handle._for_namespace("other.module")` — typo'd or deliberate — gets a fully working handle into another module's collections. `AGENTS.md` discloses this is "isolation by convention," but the method's own docstring implies a restriction the code doesn't enforce.

**6. `dumps()` uses `sort_keys=True` — silently discards dict key order despite the "lossless" docstring claim**
`libs/naas-abi-core/naas_abi_core/services/document/adapters/secondary/document_codec.py:47`
Verified via round-trip: `put(..., {"z":1,"a":2,"m":{"y":1,"b":2}})` reads back re-sorted (`['a','m','z']`, nested `['b','y']`). Any caller relying on insertion order (UI rendering, diffing) gets silently reordered output.

**7. `transaction()`'s `RLock` fully serializes reads even though WAL mode is enabled specifically to allow concurrent readers**
`libs/naas-abi-core/naas_abi_core/services/document/adapters/secondary/DocumentSecondaryAdapterSQLite.py:63`
Every `get`/`find`/`count`, read-only or not, blocks on the same adapter-wide lock. Read throughput collapses to fully serial under concurrent access, defeating the stated purpose of turning on WAL. (Note: the lock does serve a real thread-safety need for the single shared connection — it's not purely gratuitous, just over-broad.)

**8. `sqlite3.Connection.create_function()` calls run before the try/except that was added to prevent leaking the connection on setup failure**
`libs/naas-abi-core/naas_abi_core/services/document/adapters/secondary/DocumentSecondaryAdapterSQLite.py:37`
The three `create_function()` registrations sit above the `try/except BaseException: self.close(); raise` block. If any of them raises, the already-opened connection (and its WAL/lock files) leaks uncleaned — `self.close()` is unreachable from that path.

### Cleanup / duplication risks

**9. Type-ordering rank is encoded twice independently** (SQL `CASE` vs. Python if/elif) with nothing tying them together — `document_sql.py:81` and `:166` (`value_sort_parts`). A new tagged type added to `encode()`/`decode()` but only one rank table updated silently breaks keyset-pagination ordering, surfacing only as flaky pagination, not a test failure.

**10. The bytes→sortable-key transform is implemented three separate times** (Postgres SQL expression, SQLite UDF lambda, pure-Python `value_sort_parts`) — `DocumentSecondaryAdapterSQLite.py:43` and `document_sql.py`. Breaks the pattern the PR otherwise follows (equality/uniqueness centralized once in `document_codec.py`). A future encoding change is likely to update only some of the three sites, silently breaking cross-backend sort parity.

**11. `DocumentSecondaryAdapterPostgreSQL` has no `close()`**, while `AGENTS.md` documents `close()` as a general adapter capability and the SQLite adapter has one. The PR's own custom-adapter test calls `adapter.close()` in a `finally` block but only ever exercises SQLite — first real use against a Postgres-configured factory hits `AttributeError`.

**12. `DocumentAdapterConfiguration.validate_adapter()` bypasses the shared `pydantic_model_validator()` helper** that all 11 sibling `EngineConfiguration_*Service.py` files use — `EngineConfiguration_DocumentService.py:35`. Config errors surface as a bare, unlabeled `ValidationError` instead of the shared helper's descriptive, service-scoped log line.

### Efficiency

**13. Postgres adapter opens a brand-new `psycopg.connect()` per call, with no pooling** — `DocumentSecondaryAdapterPostgreSQL.py:54`. `put_many()`/`delete_many()`/`iterate()` loop over single-item adapter calls, so a bulk backfill or `delete_many()` over a few thousand documents opens thousands of fresh connections in quick succession — real risk of exhausting `max_connections` under any realistic bulk workload (a load pattern the SQLite-only local test path never exercises).

**14. `transaction(write: bool = False)`'s `write` parameter is accepted but never read on Postgres** — `DocumentSecondaryAdapterPostgreSQL.py:54`. Every call takes the same path, unlike SQLite where it selects `BEGIN` vs `BEGIN IMMEDIATE`. Harmless today, but a dead parameter a future maintainer could reasonably assume is load-bearing.

### Convention

**15. Missing `_test` files for `DocumentPort.py`, `document_codec.py`, and `document_sql.py`**. These are only exercised indirectly through `DocumentService_test.py` and the adapter contract tests — no dedicated unit test targets `validate_name`/`validate_value`/`validate_data`/`validate_query`, `encode`/`decode`/`dumps`/`equality_key`, or the SQL-compilation class by name.

---

**Process note:** `.github/workflows/pull_request.yml` never runs `make test` (only `make check` — lint/mypy/bandit — plus `make test-integration-core`, hardcoded to two unrelated triple-store files). So the ~85 new document-store tests this PR adds, **including the one that's currently failing (finding #1)**, provide no CI protection — the failure won't block merge on its own.

Dr0p42:
Addressed the review in adc548cf1, with regressions for the reported failures.

- **1:** Reproduced the quoted-field failure on SQLite 3.46 (the earlier run used 3.53). Literal top-level lookup now uses a deterministic accessor rather than JSON-path escaping. Tests cover quotes, backslashes, newlines, Unicode and tag-like keys, including unique constraints. Existing SQLite expression indexes are rebuilt transactionally on `ensure_collection`, even for unchanged declarations; duplicate data fails the repair and rolls it back.
- **2, 5:** Scoped handles retain engine service wiring and reject namespace rebinding. Only the engine-created root can bind a namespace. This guards the returned API; it does not claim to sandbox hostile code within the Python process.
- **3:** Both remote configs explicitly select PostgreSQL using deployment `POSTGRES_*` secrets. A regression renders the project scaffold and checks both configurations.
- **4:** SQLite now creates an exact equality expression index alongside the typed sort/range index. `eq` and `in` compile matching expressions; query-plan tests verify index use with representative data and statistics. Exclusions and array `contains` can still scan, which the guide now states. `indexed=True` is not a promise that every operator avoids a scan.
- **6:** Corrected the contract and wording: nested values/types round-trip, but dictionary key order is unspecified, consistent with PostgreSQL JSONB. Canonical serialization remains necessary for equality and index keys. Lists preserve order; callers requiring ordered entries should use lists.
- **7, 8:** SQLite file transactions have independent connections, with a concurrent snapshot/read/write regression. The in-memory adapter retains its serialized connection. All UDF registration is inside connection cleanup handling, with a failure-path test.
- **9, 10:** Shared value-kind constants drive SQL and cursor ranks; SQLite and Python use shared bytes sort helpers. PostgreSQL uses its native decode/hex expression. Paging tests cover every value kind in both directions on both databases, including bytes whose Base64 ordering differs from byte ordering.
- **11, 13:** PostgreSQL uses the existing psycopg pool dependency and implements `close()`, now included in the adapter protocol. Tests cover reuse, rollback, close, and initialization failure. Defaults are one initial connection, a maximum of ten, and a five-second startup/acquisition timeout; the limit and timeout are configurable.
- **12:** Configuration uses the shared service-labelled validation helper. Adapter options hide input values in validation text to avoid logging a DSN.
- **14:** Documented the shared transaction flag: it selects SQLite's deferred/immediate locking. PostgreSQL reads acquire catalog row locks and intentionally use ordinary read/write transactions; `write=False` does not imply SQL READ ONLY.
- **15 / CI:** Added direct `DocumentPort_test.py`, `document_codec_test.py`, and `document_sql_test.py` coverage. A dedicated PR job now runs `make test-document-core` on Python 3.11 and 3.12 against SQLite and a PostgreSQL 17 service.

Validation: **207 passed on each of Python 3.11 and 3.12**, using SQLite 3.53.3/3.46.0 respectively and a real PostgreSQL 17.11 instance. The full `make check` commit hook also passed (lint, core/CLI mypy, Bandit, marketplace checks, and the Nexus production build).

Dr0p42:
Follow-up validation found one more portability edge case, fixed in 9e3a409e1: JSONB can expand exponent-form floats such as `1e100` into integer text. Reading those values previously produced integers outside the port's signed-64-bit range, breaking round-trip equality, subsequent writes, and cursor validation.

Decoding now restores those out-of-range JSONB numbers to floats, recursively. Such integers cannot originate from valid integer inputs. Signed-64-bit integer boundaries remain unchanged; integral floats within that range may still normalize to integers, as documented.

Added a shared read/update/paging regression spanning large positive and negative finite floats, including the largest finite Python float. **212 tests pass on each of Python 3.11 and 3.12**, against SQLite and PostgreSQL 17.11. The full `make check` hook also passed.

Dr0p42:
## Re-review (commit `adc548c`)

New push landed since my last review, so I re-ran the full xhigh multi-angle pass against the updated diff (`origin/main...HEAD`, 33 files / ~3050 insertions). Good news first: **this commit fixes most of the previously reported issues** — confirmed by diffing `db7df35..adc548c` directly:

- ✅ #1 (SQLite field-name JSON-path escaping) — `field()` now routes through a `document_field()` SQLite UDF instead of `json.dumps()`-built paths, and equality now goes through `document_json_key()` consistently.
- ✅ #3 (`config.remote.yaml` missing `document:`) — added, routed to PostgreSQL via secrets.
- ✅ #4 (declared index doesn't accelerate equality lookups) — SQLite now builds a matching `_eq` index using the same `document_json_key()` expression the predicate compiler uses, and `ensure_index()` detects and repairs stale index definitions.
- ✅ #7 (RLock serializes all reads despite WAL) — file-backed SQLite now opens a fresh connection per transaction instead of funneling everything through one shared connection + lock; the lock only guards adapter lifecycle state now.
- ✅ #8 (`create_function()` leak before try/except) — function registration moved into `_connect()`, wrapped in its own try/except that closes the connection on failure.
- ✅ #9 (type-ordering rank duplicated in SQL vs Python) — unified behind a shared `ValueKind` enum and a single `value_sort_parts()` in `document_codec.py`.
- ✅ #2/#5 (`_for_namespace()` dropped services wiring / no access control) — now guarded by a `__can_bind_namespaces` flag only set on the engine-root instance (raises `PermissionError` otherwise), and propagates `services` wiring to the scoped instance it creates.
- ✅ #15 (missing `_test` files) — `DocumentPort_test.py`, `document_codec_test.py`, `document_sql_test.py` all added.
- ✅ CI gap — new `document-tests` job in `.github/workflows/pull_request.yml` runs `make test-document-core` against SQLite + a real Postgres service on every PR, so the test suite actually gates merges now.

New pass surfaced a fresh set of 15 findings, ranked most severe first:

### Correctness bugs

**1. `where` predicates are iterated as a `Sequence` but not materialized — a generator gets silently exhausted, turning a filtered `delete_many` into a full-collection wipe**
`libs/naas-abi-core/naas_abi_core/services/document/DocumentService.py:145`
If `where` is passed as a generator/iterator, the first internal validation pass consumes it; `document_sql.py`'s `predicates()` then sees zero clauses and falls back to `"TRUE"`, so `find`/`count`/`iterate` silently match everything — and `delete_many` deletes the whole collection instead of the intended subset. Real, unbounded data-loss risk.

**2. `ensure_collection` validates every existing document against merged field types, uncaught, inside engine boot**
`document_sql.py:302`
A module adding a `FieldSpec` type constraint for a field with pre-existing inconsistently-typed data raises `ValueError` on the next boot's `ensure_collection()` call (invoked from module `on_initialized`), with nothing catching it — the whole engine fails to boot until the data is manually migrated.

**3. Only `sqlite3.IntegrityError` is translated to the documented exception contract — everything else (e.g. `OperationalError: database is locked`) escapes raw**
`DocumentSecondaryAdapterSQLite.py:116`
A burst of concurrent writers exceeding the busy timeout raises a raw `sqlite3.OperationalError` to callers written against the documented `DocumentNotFound/CollectionNotFound/VersionConflict/UniqueViolation` contract.

**4. No upper bound on `limit` anywhere in the service/adapter/config layers**
`DocumentPort.py:170`
Only `limit >= 1` is enforced. `find("big_collection", limit=5_000_000)` builds `LIMIT 5000001` and `.fetchall()`s the whole thing in one shot — unlike `iterate()`, which caps batches internally.

**5. First-boot DDL holds a blocking advisory-lock wait inside the same transaction whose `statement_timeout` was just set**
`DocumentSecondaryAdapterPostgreSQL.py:66`
Two engines booting simultaneously against a fresh DB: if process A's DDL under the lock runs longer than process B's `statement_timeout` (default 30s), B's lock wait gets cancelled by Postgres and propagates uncaught, failing B's boot entirely.

### Correctness — edge cases

**6. Base `ensure_index` (used as-is by PostgreSQL) only does `CREATE INDEX IF NOT EXISTS`, which no-ops if an index by that name already exists — only the SQLite override detects/repairs a changed definition**
`document_sql.py:282`
A future change to index-expression generation with the same deterministic name self-heals on SQLite but leaves Postgres silently stuck on the old, semantically wrong index forever.

**7. Cursor "fingerprint" hashes `where` in the exact order supplied — reordering the same predicates invalidates a valid cursor**
`document_sql.py:459` — empirically confirmed: replaying a cursor against the same predicates in different list order raises `ValueError("Invalid cursor for this query")` even though the result set is identical.

**8. `delete(..., if_version=0)` can never succeed for any document**, unlike `put`'s well-defined create-only meaning for `if_version=0` — `document_sql.py:425`. Since a document's version is always ≥ 1, `WHERE version = 0` never matches, so it raises `VersionConflict` unconditionally instead of treating non-existence as success.

**9. `unique_together` group dedup via `dict.fromkeys` is field-order-sensitive**, so redeclaring the same compound constraint with fields reordered is treated as a new constraint — `document_sql.py:225`. A schema evolving from `('team','slug')` to `('slug','team')` creates a redundant index on every boot instead of a no-op.

### Efficiency

**10. `require_collection` (called from every `get`/`put`/`delete`/`find`/`count`) re-queries the catalog table — with a `FOR SHARE`/`FOR UPDATE` row lock on Postgres — and re-parses the spec on every single call, uncached**
`document_sql.py:184` — doubles round trips and adds lock contention for every CRUD call at scale.

**11. `EngineProxy.ServicesProxy.document` allocates a brand-new `DocumentService` on every property access, unlike every sibling service property, which returns the shared instance**
`EngineProxy.py:97` — breaks singleton-identity assumptions (`services.document is services.document` is `False` only for this service) plus unnecessary allocation on every access.

### Design / robustness

**12. `DocumentFactory` and the engine-root service both default `namespace` to the literal `"default"` with no collision guard**, even though namespace is a real storage-isolation boundary — `DocumentFactory.py:8`. Low exploitability today since real call sites pass explicit namespaces, but nothing prevents two independent callers from colliding.

**13. `IDocumentAdapter`'s `@runtime_checkable` Protocol `isinstance()` check only validates method-name presence, not signatures or the documented exception contract** — `DocumentPort.py:208`. A custom adapter with the right method names but wrong parameter order/exceptions passes boot validation silently and only fails confusingly at request time.

### Cleanup

**14. The `sqlite_equal`/`document_equal` SQLite UDF is registered on every connection but no longer referenced by any SQL the compiler emits** (equality now uses `document_json_key(...) = ?`) — `document_codec.py:113`. Dead code/unreviewed surface area.

**15. `DocumentSQL.literal()` reimplements string-literal escaping conceptually duplicated by `DatasetSecondaryAdapterDuckLake._sql_string()` elsewhere in the codebase** — `document_sql.py:58`. Not fully equivalent (psycopg `%`-doubling differs), so a future escaping fix has to be found and applied independently in both places.

---

**Heads up:** another commit has already landed on top of this one since I started this pass — I'm kicking off a fresh review against the new head now and will post it here when it's done.

Dr0p42:
Addressed this review in c35bca523. The generator-filter issue reproduced on both backends; the new shared regression verifies filtered counts, multi-page iteration, and deletion without touching nonmatching records.

- **1:** Filter iterables are materialized before validation or I/O, including direct adapter calls, and retained across iteration pages. A filter generator that raises also fails before any deletion starts.
- **2:** Retained validation during declaration/boot. Silently accepting an incompatible declaration would leave the declared type unenforced and contradict the issue's load-time failure requirement. Errors now identify the collection and offending document and direct the caller to migrate existing values. A regression verifies rollback preserves both the prior catalog and data.
- **3:** Added the portable `DocumentStorageError` contract for driver, locking, and pool failures. Both adapters translate those failures and retain the original cause. Tests cover SQLite busy timeout, PostgreSQL boot-lock timeout, pool acquisition timeout, and closed pools. A failed write can have an unknown outcome; no automatic replay is performed.
- **4:** `limit` and `batch` are bounded to **1–1,000** documents. Larger traversals use `iterate`; the SQL query fetches at most one extra continuation row.
- **5:** Retained the bounded boot-lock wait. Exceeding `statement_timeout` now produces `DocumentStorageError`, with a real contention regression. This is the documented fail-fast startup policy: orchestration can retry initialization after the other process finishes, instead of waiting indefinitely or silently extending the configured timeout.
- **6:** PostgreSQL expression indexes now carry an adapter-owned comment containing the defining SQL's fingerprint. `ensure_collection` transactionally replaces absent/stale definitions and reuses unchanged ones. Tests verify replacement, rollback when existing duplicates prevent uniqueness, and unchanged index OIDs on repeated ensure calls.
- **7:** Cursor fingerprints canonicalize AND-predicate order. Replaying a cursor with those predicates reordered now works on both backends.
- **8:** Retained the documented version precondition: a delete with any supplied version requires an existing matching document. Absence conflicts, including for zero; only unversioned deletion is idempotent for absence. Zero's create-only meaning applies to `put`. Added an explicit regression and clarification.
- **9:** Compound constraints now deduplicate field permutations while retaining the first declaration's field order and index name. Reversing the fields no longer adds a redundant constraint.
- **10:** Parsed immutable specs use a bounded **128-entry per-adapter cache**, keyed by the exact catalog JSON. The catalog query/row lock remains necessary to observe cross-process declaration changes and coordinate collection teardown. A shared regression verifies another adapter's new type constraint and collection drop are visible after cached reads.
- **11:** Each module proxy reuses its scoped service, continues checking access on every retrieval, and replaces the cached handle when the engine installs a different root. Tests cover identity, root replacement, and access revocation after caching.
- **12:** Standalone factories now require an explicit namespace. The engine root only binds namespaces and rejects storage operations, so it cannot write into an implicit default collection space.
- **13:** Kept the standard runtime Protocol check as a structural gate. It cannot certify signatures or exception semantics; that is the role of static signature checking and the shared backend contract, as the issue specifies. The guide now states this explicitly. Boot does not perform probe writes against user storage.
- **14:** Removed the unused `document_equal` registration and `sqlite_equal` helper. Equality/uniqueness tests now exercise the remaining shared key function.
- **15:** Kept the small escaping helper local to this adapter implementation. It has dialect-specific behavior, including psycopg percent escaping, and adversarial tests. Coupling it to the dataset adapter's private helper would cross service boundaries without providing an equivalent implementation.

Validation: **238 passed on each of Python 3.11 and 3.12**, with SQLite 3.53.3/3.46.0 and real PostgreSQL 17.11. The full `make check` hook passed again: lint, core/CLI mypy, Bandit, marketplace checks, and Nexus production build. The service guide and ADR document the new bounds, error contract, index upgrade policy, and retained semantics.

Dr0p42:
Re-reviewed commit **c35bca523**, including both earlier reviews and the follow-up fixes. Two remaining portability issues reproduced against SQLite and an isolated PostgreSQL 16 instance:

1. **[P2] Numeric equality and uniqueness differ between backends** — [document_codec.py:102](https://github.com/jupyter-naas/abi/blob/c35bca523909428ef28dfb045b8fbddabdb46e92/libs/naas-abi-core/naas_abi_core/services/document/adapters/secondary/document_codec.py#L102).

   Store `1.0000000000000001e18`, integer `1000000000000000128`, and integer `1000000000000000100` in separate documents. An equality filter against the float matches the float and `1000000000000000128` on SQLite, but the float and `1000000000000000100` on PostgreSQL. With a unique float field, the adapters reject opposite integer values after inserting the float. SQLite's `int(value)` canonicalizes the binary float, while PostgreSQL JSONB uses its serialized decimal representation. Both adapters need consistent numeric canonicalization across storage, equality, and ordering, with a shared regression covering these mixed float/integer values.

2. **[P2] Adding an index makes valid documents unwritable on PostgreSQL** — [document_sql.py:266](https://github.com/jupyter-naas/abi/blob/c35bca523909428ef28dfb045b8fbddabdb46e92/libs/naas-abi-core/naas_abi_core/services/document/adapters/secondary/document_sql.py#L266).

   Declare `FieldSpec(name="x", type="string", indexed=True)` and save a 5,000-character pseudorandom alphanumeric string. SQLite accepts it; PostgreSQL raises `DocumentStorageError`, caused by `index row size 5024 exceeds btree version 4 maximum 2704`. Indexing the complete scalar therefore narrows the accepted value domain even though `indexed=True` is an optimization hint. The index strategy needs to accommodate accepted values, with a regression for long, poorly compressible strings.

The earlier findings are addressed or have reasonable documented resolutions. Retaining fail-fast declaration validation, bounded startup waits, and version-checked delete semantics is appropriate.

**Validation:** all **238 existing document/engine tests passed** locally against SQLite and PostgreSQL 16. GitHub's PostgreSQL 17 jobs are also green. The additional reproductions above expose gaps in the existing coverage.

Dr0p42:
Addressed both remaining findings in 92121af7f.

1. **Numeric portability:** equality and uniqueness now canonicalize integral floats using their JSON decimal representation. The reported float matches `1000000000000000100` on both backends and remains distinct from `1000000000000000128`. SQLite numeric ranges/sorting use decimal text with a Decimal collation, so they do not round through SQLite REAL; cursor boundaries use the same decimal semantics. Reads normalize integral values where binary-to-integer conversion would change that decimal value. Shared regressions cover both signs, nested equality/array membership, uniqueness, ranges, and one-document pages in both directions. SQLite equality functions are versioned so existing indexes rebuild on `ensure_collection`; the old function remains available for safe migration and rollback. Tests cover duplicates discovered during that repair.

2. **Long indexed values:** optional PostgreSQL B-tree indexes retain the type/numeric keys and bound the text/bytes key to a 256-character prefix. Queries, equality checks, and cursors still compare full values. This preserves accepted values at the cost of potentially requiring an explicit sort for strings/bytes; GIN equality candidates and numeric/rank index prefixes remain available. Shared regressions insert poorly compressible 5,000-character ASCII and Unicode strings and long byte values, both before and after adding the index, including values sharing the entire indexed prefix. They verify equality, ranges, and ascending/descending pagination. A PostgreSQL regression also upgrades the previous unbounded index definition before writing a long value.

Validation: **253 tests passed on each of Python 3.11 and 3.12**, against SQLite 3.53.3/3.46.0 respectively and PostgreSQL 17.11. The full `make check` commit hook passed (lint, core/CLI/marketplace mypy, security checks, and Nexus production build). The service guide and ADR document the decimal contract and index tradeoff.

Please recheck these two fixes against 92121af7f; the new CI run is underway.

PR #1281, opened 2026-09-17
feat(core): Stage 1 NATS/JetStream module distribution

Dr0p42:
## What this is

First implementation slice of the distributed-modules-over-NATS RFC
(`docs/specs/rfcs/20260910_distributed-modules-nats-jetstream.md`), **Stage 1**:
engine + kernel services reachable over NATS, while every module stays fully
in-process. Modules-over-the-network (Stage 2) is separate, later work — see
the RFC's own scope note.

## What's in this PR (11 commits)

**Infra**
- NATS+JetStream added to `docker-compose.yml` and the CLI deploy template,
  opt-in alongside RabbitMQ/Redis, not a replacement yet.
- `abi dev up --service nats` — a native `nats-server` process (opt-in only,
  never part of the default `abi dev up`), following the exact same pattern
  `oxigraph` already uses for a locally-managed dev dependency.

**Bus**
- `NATSJetStreamAdapter implements IBusAdapter`: core NATS pub/sub for
  `publish`/`subscribe` (per-subscriber fanout), a JetStream stream + durable
  pull consumer per `(topic, routing_key)` for `enqueue`/`dequeue`
  (competing-consumer, exactly-once). Wired as `bus_adapter: "nats_jetstream"`.
- This one is a **complete, working, end-to-end path today** — pub/sub is
  peer-to-peer through the broker, no separate "server" needed.

**Kernel service RPC — every service now has a NATS adapter pair**
Started with `object_storage` as the pilot, then extended the exact same
pattern to every other kernel service: `dataset`, `triple_store`,
`vector_store`, `kv`, `email`, `cache`, `activity_log`, `event`,
`coding_environment`, `source_control`, and — at Max's explicit direction,
overriding the earlier "skip for security reasons" call — `secret` too
(12 services total).

For each: a hand-designed Protobuf contract under `naas_abi_core/proto/<domain>/v1/`
(sharing the `CallContext`/`CallError` envelope from `common.proto`, now with
an `optional int32 status` field for exceptions that carry an HTTP-style
status code), a `<domain>_nats_contract.py` neutral module (wire constants
neither adapter owns), a primary adapter (server) and a secondary client
adapter (symmetric error-code → exception mapping), and `nats_rpc` wired into
each service's existing `adapter: Literal[...]` config slot alongside
`fs`/`redis`/`sqlite`/etc.

One architectural rule applied uniformly, derived from how `object_storage`
already worked: **the NATS client implements the domain's secondary
port**, plugged into the same config slot every other concrete adapter
uses; **the NATS primary wraps the domain SERVICE** (not the bare adapter)
wherever the domain layers real behaviour on top — event publishing,
derived checks — so remote callers get identical behaviour to in-process
callers, not a divergent code path. A few services deviate deliberately:
`event`'s primary/client only cover the 6-method durable-log port
(`IEventAdapter`) — `publish`/`subscribe`/bus broadcasting stay 100% local,
composed from those RPCs — and `vector_store`/`cache` wrap the raw port
directly since there's no richer domain-level side effect at that specific
boundary. `cache` is also the one service **not** wired into
`EngineNATSLoader`: it's multi-tier (`CacheService` orchestrates one or
more `ICacheAdapter`-backed tiers), so its NATS client is just one more
per-tier adapter choice in the existing config list, already usable
through `config.yaml` with no loader change. `secret` fans out over a
*list* of adapters rather than wrapping exactly one, so its NATS client
plugs in as one more `secret_adapters` list entry (not a top-level swap),
and its `EngineNATSLoader` re-exposure guard is `all(...)`, not a single
`isinstance` check — skipped only when *every* configured adapter is
itself a NATS client, since a local dotenv adapter alongside a remote one
still has something real worth serving.

Deliberately **not** ported, flagged rather than silently skipped:
- `model_registry` — has no secondary-adapter-port to hang a NATS client on
  at all (its config builds the service straight from two strings, no
  `Literal[..., "custom"]` slot), and its `get*` methods return live
  LangChain client objects bound to local credentials — fundamentally
  process-local, not serializable.

Streaming/callback members explicitly excluded per service, same treatment
as `object_storage`'s `get_object_stream`/`put_object_stream`: `keyvalue`'s
`lock()`, `coding_environment`'s `wait_until_ready`, `event`'s
`subscribe`/`iter_query*`, `cache`'s memoization decorator — all stay
client-side, composed from the now-remoted primitives, no new RPC needed.

**`secret`'s security note, carried in its own docstrings, not glossed
over**: Stage 1's shared-JWT auth has no per-caller ARN/IAM authorization,
so any process holding a valid service token can read/write/list every
secret the remote process's `Secret` service exposes. Max's call,
explicit and recorded: *"Yes for now, it's fine. If it's adding security
problems we will have to fix that anyway."* Revisit once Stage 2
per-caller authorization exists.

**Engine-boot exposure**
- `EngineNATSLoader.expose_services()` (wired into `Engine.load()`) now
  auto-starts a primary adapter for all 11 wired services the moment
  `config.yaml`'s `nats:` block is present — the same "presence alone turns
  it on, no per-service flag" design as the `object_storage` pilot, just
  extended. Its test suite is a generic, parametrized table covering all
  11 (noop cases, one-primary-per-service, re-exposure guard, and a
  combined "everything loaded at once" boot scenario — 29 tests).
- `nats_runtime.py`: unchanged, still the one shared lazily-connected NATS
  client per process every primary adapter here registers against.

**Graceful shutdown**
- `Engine.shutdown()`: stops every NATS primary adapter this engine's
  `load()` started (draining its NATS micro-service registration while the
  connection is still up), then closes the shared `nats_runtime`
  connection. No-op if `nats:` was never configured; safe to call more
  than once or before `load()` ever ran; a primary that fails to stop is
  logged and skipped rather than blocking the rest of shutdown.
- Wired into `apps/api/api.py` via a FastAPI `lifespan` context manager —
  there wasn't one before at all. Reads `LazyEngine`'s private `_engine`
  directly (not through its `__getattr__` proxy) so a shutdown on a
  process that never actually loaded the engine doesn't force one into
  existence just to tear it down.
- Known, accepted residual gap: `apps/dagster/dagster.py` also boots a
  full `Engine()` but has no natural "about to exit" hook the way
  FastAPI's lifespan does (worker processes are managed externally).
  `nats_runtime.close()` is already `atexit`-registered, so the shared
  connection still closes there — only the primary adapters' *graceful*
  `.stop()` is still missing for that specific entry point. Revisit once
  Dagster's own calls actually go through these NATS-exposed services (the
  RFC's own next step, not done yet).

## What's still NOT done

- `model_registry` port, for the reason above — tracked as real future
  work, not an oversight.
- Tracked in the RFC, not blocking this PR: the ontology-bootstrap-ownership
  KV flag, and pointing Dagster's own calls at these NATS-reached services
  (see the graceful-shutdown residual gap above, which depends on this).

## Verification

Every service verified against a **real running `nats-server -js` process**
(not mocks, not testcontainers-in-CI — this environment has no Docker
daemon, confirmed explicitly, so the Docker-backed integration tests in each
adapter's own test suite self-skip here by design and are additionally
covered by this direct verification): a real primary adapter wrapping a
real domain service/adapter (filesystem, SQLite, in-memory, embedded
Qdrant, embedded Oxigraph — whatever each service's own lightest real
backend is), a real client adapter making real network round trips through
it, asserting correct results **and** the correct domain-exception round
trip on at least one error path per service (`ObjectNotFound`,
`DatasetNotFoundError`, `KVNotFoundError`, `RepoNotFoundError`, auth
rejection). All 12 services (11 wired + `cache` standalone) pass live.
Graceful shutdown separately live-verified: started a real primary
adapter, confirmed a separate NATS connection could reach it, ran the
shutdown sequence, then confirmed that same connection gets a genuine
`NoRespondersError` afterward — not just "the process didn't crash."

Full test suite: ruff clean, mypy clean (0 errors across 657 source files),
and every touched service's own test suite green (unit tests for the
primary adapter's auth/dispatch/error-mapping, the client adapter's
lazy-connect/token-reissue/error-mapping, and — where the domain already
had one — the shared generic secondary-adapter contract test, parametrized
in). Full pre-commit hook (ruff via `uvx`, mypy per-package, bandit) passes
on every commit.

**Known pre-existing, unrelated issues confirmed (not introduced by this
change — reproduced identically with zero diff on the files involved)**:
the 9 `engine_configuration/`/`Engine_test.py` failures already documented
on this PR; a genuine latent bug in
`TripleStoreService__SecondaryAdaptor__Filesystem` where schema bootstrap
unconditionally inserts into a named graph that adapter explicitly doesn't
support (`NotImplementedError`); a genuine latent bug in
`DotenvSecretSecondaryAdaptor.list()`, which reads a construction-time
snapshot and never reflects a `set()` made afterward (`get()` happens to
still work, via its separate `os.environ` fallback); two
`EngineConfiguration_SecretService_test.py` failures needing local
`.env`/`NAAS_API_URL` setup this environment doesn't have; three more
`EngineConfiguration_test.py` failures from macOS's `/tmp` → `/private/var`
symlink plus test-order-dependent state, both already known repo quirks;
one `NaasSecret_test.py` failure that depends on live network access and
passes fine in isolation.


Dr0p42:
## Pre-merge checklist

Notes from a review pass so they don't get lost.

- [ ] **Resolve conflicts with `main`** (52 commits behind). All three are trivial:
  - `libs/naas-abi-core/naas_abi_core/apps/api/api.py`: import clash (`import html` vs `import asyncio`). Keep both.
  - `libs/naas-abi-cli/naas_abi_cli/cli/deploy/local_test.py`: both sides appended tests (admin-password tests from main, NATS `max_payload` test here). Keep both.
  - `libs/naas-abi-cli/uv.lock`: version numbers only. Take main's and re-lock.
- [ ] **GitGuardian is a false positive.** Incident 35674542 flags `demo_password="s3cret"` in `libs/naas-abi/naas_abi/agents/tools/apps_tools_test.py:54`, a test fixture that came in through the merge of `main` (f65297183), not from NATS code. Mark it as a false positive in the dashboard.
- [ ] **Lock down NATS before merge.** Both `docker-compose.yml` and the CLI deploy template publish `4222:4222` and `8222:8222` on all interfaces, and `nats.conf` only sets `max_payload`: there is no auth. With the `nats` profile enabled, anyone who can reach the host can call every kernel service over RPC, including reading secrets via `secret`, minting Forgejo tokens via `source_control.MintGitToken`, and reading/writing the triple store and object storage. Minimum fix:
  - bind the ports to `127.0.0.1` (or drop the host mappings, the services talk over `abi-network`);
  - add token or user auth in `nats.conf` and have the engine's `nats:` config pass it.

  The RFC's shared-JWT model can come after.
- [ ] **`CallContext` identity is not implemented.** The RFC says `CallContext` carries `principal_id` / `workspace_id` / `tenant_id` starting in Stage 1 (carried, not enforced). Nothing on this branch sets or propagates it. Either add it here or update the RFC to say it is deferred. It is the foundation for per-user authorization over NATS.
- [ ] **Update the PR description.** It still says "11 commits", the branch now has 19.

PR #1299, opened 2026-09-22
feat: standalone NATS modules, domain boundaries, and document checkpoints

Dr0p42:
Modules running outside the engine previously needed ABI core, and internal domain dependencies still used local calls. This change adds independent proto/SDK libraries, a familiar remote module lifecycle, and network-only domain dependencies when top-level NATS is configured. The runnable module exercises engine services exclusively over NATS from a separate process.

This PR is stacked on #1281 (`feat/nats-jetstream-stage1`) so the diff contains only this next step. Merge #1281 first, then retarget this PR to `main`. This branch also merges current `main` (including the document service).

## Changes

- **naas-abi-proto:** canonical schemas, generated messages, and type stubs; depends only on protobuf. Existing core imports forward to these classes, preserving wire contracts. Core includes it through its optional NATS extra.
- **naas-abi-sdk:** typed async clients for all 109 existing v1 RPC endpoints, sharing one connection. Depends only on proto and nats-py; accepts issued service tokens without requiring the signing key. Preserves bounded payloads/deadlines and does not automatically replay failed RPCs. Includes native pub/sub, batch publication, and durable queue helpers.
- **Remote module lifecycle:** SDK `BaseModule`, nested `Configuration`, `ModuleDependencies`, `engine.services`, and load/initialize/unload hooks. `run_module` / `naas-abi-module` owns transport and cleanup. Undeclared service access and unsupported module discovery fail explicitly.
- **NATS domain boundaries:** retain private domain owners and inject NATS-backed service views into domains and local modules. Endpoints wrap owners, never proxies. Calls fail without an in-process fallback. Project NATS also selects the native bus; explicit upstream NATS service routes are retained. Each primary has independent worker capacity so nested cross-domain calls do not exhaust a shared executor.
- **Cache composition:** optional KV-backed hot cache and object-storage-backed cold cache, loaded transitively and accessed through NATS. Preserve ordered tiers with additional tier subjects. Service-level endpoints emit events once.
- **Document service over NATS:** eight typed storage primitives, lossless portable values, CAS writes, paging, scoped module clients, upstream `nats_rpc` configuration, and complete adapter-contract coverage over a real broker.
- **Optional LangGraph checkpointer:** `naas-abi-sdk[langgraph]` adds async document-backed checkpoints, history, pending writes, interrupt/resume and deletion. A second interpreter resumes an interrupted graph from shared documents. Base SDK dependencies stay unchanged; existing core Agent/PostgreSQL configuration and stored history are untouched.
- **Remote-agent design:** proposed discovery/readiness, invocation/status/event contracts, and execution ownership in `docs/specs/rfcs/20260922_remote-agent-invocation.md`. This does not yet implement remote agent calls.
- **First standalone module:** `examples/standalone_module` starts a real engine and temporary JetStream broker, then runs a worker in a fresh environment containing only the two libraries, protobuf, and nats-py. The worker extends SDK `BaseModule`, asserts core is absent, and reports process IDs, lifecycle, and operation coverage. The host observes real cache-to-KV/object-storage/event broker traffic and checks mutation events are not duplicated, with the shared asyncio pool restricted to one worker.
- **Required engine fixes:** expose the existing cache RPC adapter for the cold tier; keep dotenv reads/listing and persistent removal consistent after mutations, as uncovered by the demo.
- Add an ADR, package Makefiles, generation instructions, and an independent Python 3.10/3.11/3.12 CI matrix.

## Review and run

Start with `libs/naas-abi-sdk/README.md`, `examples/standalone_module/worker.py`, `libs/naas-abi-core/naas_abi_core/engine/engine_loaders/EngineNATSDependencies.py`, and the two `20260922` standalone SDK / NATS domain boundary ADRs.

With the repository development environment and `nats-server` available:

```sh
make demo-sdk
make test-sdk
```

The demo prints results and writes `examples/standalone_module/report.json`. It verifies object create/read/list/update/metadata/delete, corresponding mutable-service lifecycles, log queries/cursors, and bidirectional bus traffic. Data and broker processes are temporary; email is saved locally.

## Validation

- 109 RPC endpoints exercised across 13 services, plus five bus methods; no missing exercises.
- Worker installed exactly four packages, with a different PID from the engine and no ABI core installed.
- 14 SDK tests, 209 engine regressions, 225 primary-adapter tests, and six native NATS integration tests passed.
- Observed cache hot/cold calls crossing the broker into KV and object storage; verified each underlying mutation event is emitted once.
- Document follow-up: 256 document/checkpoint tests passed (77 PostgreSQL tests skipped without a configured database); final transport/graph rerun passed 72 tests. Optional isolated SDK passed 18 tests; engine wiring passed 52 tests; configuration/SDK follow-up passed 34 tests.
- Full commit checks passed: lint, security scanning, core/CLI/marketplace type checks, and Nexus build.
- Built wheels and verified all 14 schemas and generated type stubs are included.

## Limits and release notes

Projects without top-level `nats` retain local wiring and optional NATS imports. Enabling it intentionally changes cross-domain traffic, bus selection, and failure behavior. Mixed local/remote secret fanout is rejected to avoid recursive self-routing. Model registration remains process-local for module bootstrap and is excluded from domain dependencies.

Migration preserves module structure, not arbitrary module binary compatibility: change imports/dependency declarations, use dataclass configuration, and await protobuf service calls. Framework component discovery is not included. Multiple engine processes still require explicit ownership/deployment; this PR does not add placement or ownership leases.

The triple-store internal `handle_view_event` endpoint returns its expected `NOT_SUPPORTED` response and is recorded explicitly. Object streaming, process-local model objects, and agent/ontology execution have no v1 remote contract. Coding/source-control use the engine's in-memory providers; external providers are not validated by this demo. This does not implement module registration, heartbeat, remote agent invocation, or workload scheduling. The new checkpointer supports async LangGraph calls only, uses full snapshots within the existing payload limits, requires one active run per agent/thread, and does not migrate PostgreSQL checkpoints.

Authorization remains the existing Stage 1 shared-trust model, with broker credentials/subject ACLs separate from service JWTs. Existing v1 schemas still have no Protovalidate annotations. Remote administrative shutdown endpoints are exercised only against the disposable demo engine.

Both new packages start at 0.1.0 and are unpublished. Publish proto before releasing core/SDK versions requiring it; local UV source overrides support development meanwhile. Remote CI results are separate from the local validation above.


libs/naas-abi-core/naas_abi_core/engine/engine_loaders/EngineNATSDependencies.py
```diff
@@ -0,0 +1,242 @@
...
+    def remove(self, triples: Graph, graph_name: URIRef) -> None:
+        return self.adapter.remove(triples, graph_name)
+
+
+class EngineNATSDependencies:
+    def __init__(self, config: NATSConfiguration):
+        self.config = config
+        self.clients: list[Any] = []
+
+    def _client(self, cls, **kwargs):
```
Dr0p42: **[Blocker] Every engine call to a domain now runs one at a time.**

In NATS mode, in-process module calls and owner-to-owner calls all go through a single `NatsRPCClient` per domain, and `NatsRPCClient._call` holds `self._call_lock` for the whole request/reply (`nats_rpc.py:232-240`).

Example: 20 concurrent FastAPI requests reading object storage (or KV, or events) used to run in parallel against the local adapter. Now each one waits on this client's lock. Nested chains make it worse (cache → object_storage facade → owner publishes an event → events facade). The inner calls queue behind unrelated callers and use up the outer call's 10s timeout, so callers get `DEADLINE_EXCEEDED` even when the operation completed.

Suggest not holding the lock across the request/reply (nats-py's `request` is safe to call concurrently), or using a small client pool per domain.

libs/naas-abi-core/naas_abi_core/services/model_registry/adapters/primary/model_registry_nats.py
```diff
@@ -0,0 +1,418 @@
...
+        self.max_payload = 512 * 1024
+        self._active = 0
+        self._reaper: asyncio.Task | None = None
+
+    async def start(self, nc) -> None:
+        self.max_payload = min(nc.max_payload, self.max_payload)
+        try:
+            for operation in OPERATIONS:
+                self.subscriptions.append(
+                    await nc.subscribe(
```
Dr0p42: **[Blocker] No queue group, so every engine on the broker handles every model-registry request.**

These endpoints use a plain `nc.subscribe`. The api and dagster processes both boot `Engine()` from the same `config.yaml` with `nats:` set, so:
- chat and embed inference runs, and is billed, twice;
- `stream_open` creates two streams, and the caller keeps whichever id answers first. The other stream is orphaned and keeps generating;
- on `stream_next`, the engine that doesn't own the stream raises `KeyError` and replies `NOT_FOUND`, often before the owner has a chunk, so remote streams fail at random.

Stateless ops need a `queue=` group. Streams need subjects routed to the engine that owns the stream, for example by including an engine id in the stream subject. `discovery_nats.py:42` has the same issue, so registry changes are also applied twice.

Dr0p42 (commented):
Review of the NATS SDK changes. Four blockers (inline, marked **[Blocker]**):

1. Every engine call to a domain runs one at a time, because `NatsRPCClient._call` holds its lock for the whole request/reply. This causes false `DEADLINE_EXCEEDED` under concurrency.
2. Model-registry (and discovery) subscriptions have no queue group, so every engine on the broker handles every request. Inference runs twice and remote streams break.
3. Cache tiers backed by NATS are listed in `describe` but never served.
4. The configured `bus` section is ignored in NATS mode (`emit_message_events` and the adapter choice are dropped silently).

The other comments cover reliability, performance and test coverage.

libs/naas-abi-core/naas_abi_core/engine/engine_loaders/EngineNATSLoader.py
```diff
@@ -312,8 +331,52 @@ def expose_services(self, services: IEngine.Services) -> list[object]:
...
             )
 
+        if services.cache_available() and not isinstance(
+            services.cache.cold.adapter, CacheSecondaryAdapterNATSClient
+        ):
+            # v1 exposes one adapter. Use the canonical cold tier for remote callers.
+            primary_cache = CachePrimaryAdapterNATS(
+                services.cache.cold.adapter,
+                nats_config.jwt_secret,
+                tiers=tuple(tier for tier, _ in services.cache.adapters),
```
Dr0p42: **[Blocker] Cache tiers backed by NATS are listed in `describe` but never served.**

`describe` lists every configured tier (`tiers=tuple(tier for tier, _ in services.cache.adapters)`), but the per-tier loop skips tiers backed by a NATS client, so those indices get no `abi.svc.cache.v1.tier.<index>` endpoint.

With tiers `[('hot', nats_rpc), ('cold', fs)]`, `tier.0` is never served. The SDK's `CacheService._views()` calls `tier.0` for get/exists/delete and either gets a `NoRespondersError`, which isn't caught because only `RPCError` is caught, or reaches another engine's tier 0. The tier subjects are global, not scoped per engine. Result: failed or wrong cache reads, and deletes that miss entries.

libs/naas-abi-core/naas_abi_core/engine/engine_loaders/EngineServiceLoader.py
```diff
@@ -56,6 +56,15 @@ def _should_load_service(
...
         return service_type in reachable
 
+    def _load_bus(self) -> BusService:
+        if self.__configuration.nats is None:
+            return self.__configuration.services.bus.load()
+        from naas_abi_core.services.bus.adapters.secondary.NATSJetStreamAdapter import (
+            NATSJetStreamAdapter,
+        )
+
+        return BusService(NATSJetStreamAdapter(self.__configuration.nats.nats_url))
```
Dr0p42: **[Blocker] The configured `bus` section is ignored in NATS mode.**

In NATS mode this builds a bare `BusService(NATSJetStreamAdapter(nats_url))`, and so does the dependency facade (`EngineNATSDependencies.py:115-117`). Adding a top-level `nats:` block to a deployment that sets `services.bus.emit_message_events: true` or uses the rabbitmq adapter stops the `BusMessagePublished`/consumed audit events and moves traffic off RabbitMQ, with no warning or config error.

Suggest either honouring `services.bus` (including `emit_message_events`) or failing config validation when the two conflict.

libs/naas-abi-sdk/naas_abi_sdk/agent_host.py
```diff
@@ -0,0 +1,443 @@
...
+            run.data, run.version = candidate, doc.version
+
+    async def _event(self, run: _Run, event: dict) -> None:
+        if run.context.cancelled.is_set():
+            raise asyncio.CancelledError()
+        if (
+            set(event) != {"event", "data"}
+            or not all(isinstance(v, str) for v in event.values())
+            or len(event["event"]) > 64
+            or len(event["data"].encode()) > 64 * 1024
```
Dr0p42: **Remote agent streams fail on normal-sized runs.**

The limits here are 64 KiB per event and 1024 events per run. `RemoteAgentAdapter` forwards core `Agent.stream_invoke` events unchanged, including `call_model` events whose `data` is `str(message.payload)`, which holds the full message state.

With a long conversation or large tool results, one event goes over 64 KiB, `_event` raises `OUTPUT_LIMIT`, and the whole invocation is marked FAILED even though the agent produced a valid answer. Long tool-heavy runs hit the 1024-event cap the same way. Consider truncating or summarising oversized `data` instead of failing the run, and making the event cap configurable or ring-buffered.

libs/naas-abi-core/naas_abi_core/engine/engine_loaders/EngineNATSDependencies.py
```diff
@@ -0,0 +1,242 @@
...
+            )
+            self.clients.append(models)
+            values["model_registry"] = models
+        dependencies = IEngine.Services(**values)
+        if dependencies.triple_store_available():
+            dependencies.triple_store.set_services(
+                IEngine.Services(bus=values.get("bus"))
+            )
+        # These two primaries expose raw adapters, so the client facade owns events.
+        for key in ("cache", "vector_store"):
```
Dr0p42: **Audit events depend on where the caller runs.**

Because the cache and vector_store primaries expose raw adapters and only this client facade emits events, an SDK module calling `cache.set_json` or a vector_store upsert over NATS writes straight to the adapter with no CacheSet/vector events. The same write from an in-engine module does emit them. It would be more consistent to emit events on the primary (owner) side.

libs/naas-abi-sdk/naas_abi_sdk/discovery.py
```diff
@@ -0,0 +1,316 @@
...
+                                uuid4().hex,
+                            )
+                            await self.register()
+                    except (RPCError, NATSError, asyncio.TimeoutError, OSError):
+                        logging.getLogger(__name__).warning(
+                            "Discovery re-registration failed", exc_info=True
+                        )
+                elif exc.code in ("UNAVAILABLE", "REGISTRY_BUSY"):
+                    self.status = "UNAVAILABLE"
+                else:
```
Dr0p42: **A healthy module can shut itself down on an unexpected error.**

The heartbeat re-raises any `RPCError` other than `LEASE_EXPIRED`/`UNAVAILABLE`/`REGISTRY_BUSY`, and any non-transport exception from `on_registered` (`AgentHost._bind`). That stops lease renewal for good, `run_module`'s `asyncio.wait` sees `registration.task` done, and the process shuts down.

Example: `UNAUTHENTICATED` during a token-secret rotation, `PAYLOAD_TOO_LARGE`, or a subscribe error in `_bind` when re-registering after `LEASE_EXPIRED`. Suggest logging and retrying with backoff for everything except clearly fatal codes.

libs/naas-abi-sdk/naas_abi_sdk/agent_host.py
```diff
@@ -0,0 +1,443 @@
...
+            response.error.code, response.error.message = (
+                "PAYLOAD_TOO_LARGE",
+                "Agent response exceeds limit",
+            )
+            payload = response.SerializeToString()
+        if msg.reply:
+            await msg.respond(payload)
+
+    async def _view(self, data: dict, after: int = 0) -> pb.Invocation:
+        instances = await self.session.client.get_module(
```
Dr0p42: **Registry load grows with every waiting caller.**

`_view` runs a discovery `get_module` RPC on every submit/status/cancel just to compute `owner_available`, and `InvocationHandle.events()` polls status every 100ms. That is about 10 registry reads per second per waiting caller, and each read fetches and parses the full JetStream KV snapshot (up to 512 KiB). With N concurrent invocations this slows every other discovery operation. Caching the membership view for each lease interval would avoid it.

libs/naas-abi-core/naas_abi_core/services/model_registry/adapters/primary/model_registry_nats.py
```diff
@@ -0,0 +1,418 @@
...
+            stream = self.streams[request.stream_id]
+            if stream.caller != caller:
+                raise PermissionError()
+            if stream.reading or request.sequence != stream.sequence:
+                raise ValueError("Concurrent or out-of-order stream read")
+            stream.reading = True
+            stream.touched = asyncio.get_running_loop().time()
+            try:
+                while stream.queue.empty() and not stream.task.done():
+                    await asyncio.sleep(0.01)
```
Dr0p42: **`stream_next` busy-waits.**

`while stream.queue.empty() and not stream.task.done(): await asyncio.sleep(0.01)` wakes the shared `nats_runtime` loop 100 times a second per waiting stream. With up to 32 streams waiting on a slow provider's first token, that wastes CPU and adds latency jitter to every other NATS endpoint on the loop. Awaiting `queue.get()` raced against `stream.task` (`asyncio.wait(..., return_when=FIRST_COMPLETED)`) would avoid it.

libs/naas-abi-core/naas_abi_core/services/model_registry/adapters/primary/model_registry_nats.py
```diff
@@ -0,0 +1,418 @@
+"""Authenticated model lookup/inference boundary with bounded pull streams."""
```
Dr0p42: **Missing colocated tests.** `model_registry_nats.py`, `model_registry_client.py`, `discovery_nats.py` and `discovery_jetstream.py` ship without `_test` files next to them. The stream, deadline and error mapping here and the JetStream registry's compare-and-swap logic would benefit most from unit tests.

